PR #47, opened 2023-06-16
Resolving a bug.

dmkhitaryan:
Providing a scale argument leads to the case where the lines 215-218 led to RTs being too small and thus all rejected, leaving 0 epochs. I commented out those lines and the code runs as intended.


GWeindel (commented):
This is more a documentation error and a bad example in the tutorial. I'll reject this PR because this is something one can need if e.g. the RTs are in ms scale, but we will be sure to properly document that parameter and not include it in the tutorial to avoid confusion

PR #49, opened 2023-06-19
Added section on already stim-resp epoched data

jelmerborst:
Added a section on already preprocessed and stim-resp epoched data and made some changes to the text.


GWeindel (changes requested):
Thanks Jelmer. However regarding the new section you suggest, the notebook would fail to execute if the data is not in the specified folder, I would suggest either to 1) provide a sampled data as is done for the other methods or alternatively 2) to comment out the lines to avoid breaking the notebook's execution. 
Same is true for the last section of the tutorial 1 but at least we hint readers how to get the data

GWeindel (approved):
Thanks looks like it works now. FYI at some point we should probably make the data downloadable rather than having it all shipped when cloning the repo

PR #51, opened 2023-06-19
Three options for zscoring in transform_data

jelmerborst:
In transform data, you can now opt for zscoring across all data (matlab default), by subject (preferred), or by trial. If setting parameter to True, it will zscore by trial, as before.


GWeindel:
Thanks, Jelmer for this PR and the other one. However, these PR aren't based on the latest version of the main branch. Can you pull/rebase and resuggest these PR with your branch up-to-date? I can also work on the conflicts but there are quite a few so it'll take some time

PR #52, opened 2023-06-19
Fix for changing time_step in plot_topo_timecourse

jelmerborst:
When changing step without explicitly setting the RT to be plotted, the RT would be plotted multiplied by the time_step. Ie when using a timestep of 10 and mean RT was 8000 ms, it would be plotted at 80,000 ms, as correction was applied twice.

### DIFF
--- a/src/hsmm_mvpy/visu.py
+++ b/src/hsmm_mvpy/visu.py
@@ -4,27 +4,26 @@
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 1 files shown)


GWeindel:
Thanks, Jelmer for this PR and the other one. However, these PR aren't based on the latest version of the main branch. Can you pull/rebase and resuggest these PR with your branch up-to-date? I can also work on the conflicts but there are quite a few so it'll take some time

PR #54, opened 2023-06-20
zscoring by subject/trial/all data

jelmerborst:
added the option to zscore either by subject, trial, or across all data. Should be backward compatible.

### DIFF
--- a/src/hsmm_mvpy/utils.py
+++ b/src/hsmm_mvpy/utils.py
@@ -485,26 +485,26 @@ def stack_data(data, subjects_variable='participant', channel_variable='componen
     return data #xr.Dataset({'data':data, 'durations':durations})
     #return data, durations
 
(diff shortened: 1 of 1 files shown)


GWeindel (changes requested):
Looks fine, only two things:
- 'subject' is used in the PR instead of 'participant' as in the rest of the code, so to remain consistent 'participant' should be used as a valid string
- can you also verify that the ordering of the dimensions in the xarray is the same after these different options? I had troubles with that in the past when stacking and unstacking

jelmerborst:
I changed subject to participant, also in the LOOCV functions. Note that we do use a subj_idx in other functions, but they are named so I did not change those. I also made sure the dimensions are the same for all zscoring options.

GWeindel (approved):

PR #55, opened 2023-06-20
note on mp in vscode on mac in tutorial 2

jelmerborst:
multiprocessing in an interactive window in VS code on Mac created strange behavior due to continuously rerunning code. This can be fixed by setting the start method of mp to 'fork' (default is 'spawn'). I added a note in the tutorial explaining how to fix this if you run into this specific issue. Given that the 'spawn' method will become the mp default across platforms, I assume this will be fixed at some point in VS code, which is why I didn't adjust anything in HMP code base.

### DIFF
--- a/tutorials/2-Estimating_a_model.ipynb
+++ b/tutorials/2-Estimating_a_model.ipynb
@@ -1,6 +1,7 @@
 {
  "cells": [
   {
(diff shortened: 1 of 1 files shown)


GWeindel (approved):
Thanks!

PR #58, opened 2023-06-22
changes in tutorial 2/models/visu

jelmerborst:
Changed some small textuals in tutorial 2.

Removed bug in models.py, in EM. Fix to remove float imprecision introduced division by 0 and ensuing problems. Now it works fine directly, also finds sensible topos with only a single iteration. Made some additional small changes to locations.

Changed the plot_topo_timecourse to keep topos at correct location when magnifying. Middle of topo is now located on middle of event. I also added potential lines to indicate this time point (event_lines), but that only works well when magnify => 2. We should probably indicate area of event (from start to end) instead with a light color behind topos.


GWeindel (approved):
Nice PR, everything looks good to me!

Issue #63, opened 2023-06-26
Contribution of the channels to the transition events (magnitudes) are now computed on the maximum probability at each trial


GWeindel:
Up to now the contribution of each channel to the transition event is computed as follows:

```python
for comp in range(self.n_dims):
    magnitudes[event,comp] = np.mean(np.sum( \
        eventprobs[:,:,event]*self.data_matrix[:,:,comp], axis=0))
# Scale cross-correlation with likelihood of the transition
# sum by-trial these scaled activation for each transition events
# average across trials
```

But averaging the probabilities in this way : 
1. slows down the estimation process as the mean position is more influenced by the gamma distribution
2. forces us to create a new data representation (see Issue #62 )
3. Based on simulations seems to actually misestimate the event timings (see last point)

A new proposition is to base the computation of the contributions to the max probabilities
```python
if self.em_method == "max":
    #Take time point at maximum p() for each trial
    #Average channel activity at those points
    event_values = np.zeros((self.n_trials, self.n_dims))
    for trial in range(self.n_trials):
        time = self.starts[trial]+ np.argmax(eventprobs[:, trial, event])
        event_values[trial] = self.events[time]
    magnitudes[event] = np.mean(event_values, axis=0)
```

Latest commits in Branch #Improving-EM-estimation by default uses the max method as it seems the best both in terms of correctness and speed:

![image](https://github.com/GWeindel/hsmm_mvpy/assets/28750455/7db5c65b-c96e-4409-9a80-0edc0383690e)
![image](https://github.com/GWeindel/hsmm_mvpy/assets/28750455/fc54a619-89e4-47a1-a38d-f5f9c51b3e21)

GWeindel:
This also has as a consequence that backward_estimation() yields more coherent event repartition, example on tutorials 3 data:

with the max method:
![image](https://github.com/GWeindel/hsmm_mvpy/assets/28750455/a4e69682-75a4-44f3-aec1-6f95e91ce1b3)
With the mean method
![image](https://github.com/GWeindel/hsmm_mvpy/assets/28750455/b1cf45c0-de9f-408b-ab3a-954c65775791)

Issue #62, opened 2023-06-26
Re-organizing the cross-correlated values is now done at the init phase instead of repeatedly on each EM() call


GWeindel:
__Current behavior__:
At each call of ```EM()``` the crosscorrelated data (```all_samples X n_dimensions```) in ```self.event``` is re-organized into a by-trial matrix (```max_samples X n_trials X n_dimensions```).

__Problem__: Given that the cross-correlated data is not susceptible to change once ```hmp.models.hmp()``` is initialized, the most efficient way is to do it only once at the init call. 

__Consequence__: speed up is very modest for single call of ```EM()``` but function making repeated calls (```hmp.fit(), hmp.sliding_event, hmp.backward_estimation```) the speed up can be as high as x2

__possible problems__:
- If crosscorrelated data is changed after init by modifying ```self.events```, it will not be taken into account but this is not a behavior that should be done.
- size of the initialized model increases but shouldn't be a problem given usual data size

Will be shipped with issue #61

Issue #61, opened 2023-06-26
Implementing a minimum number of iterations in the Expectation maximization algorithm.


GWeindel:
__Current behavior__: the EM stops as soon as the likelihood from a new iteration is lower than the previous iteration (```while lkh - lkh_prev > threshold```):
```python
        while lkh - lkh_prev > threshold and i < max_iteration:#Expectation-Maximization algorithm
            #As long as new run gives better likelihood, go on 
```

__Problem__: with bad starting points of $\theta$ and scarce data it can happen that the while loop is broken without having reached convergence (stablization of the loglikelihood).

__Solution__: A minimum number of iterations fixes this behavior
Example with data from tutorial 2
![image](https://github.com/GWeindel/hsmm_mvpy/assets/28750455/3ff8c500-8f00-418d-bd90-39281cab25fa)
__Proposed implementation__ in branch #Improving-EM-estimation
When calling ```hmp.models.hmp()``` a minimum number of EM iterations can be specified. 
```python
        while i < max_iteration :#Expectation-Maximization algorithm
            if i > min_iteration and lkh - lkh_prev < threshold:
                break
```
The default is 10 and this can be overriden for specific fits when calling the ```EM()``` function if needed (e.g. the last step of the ```fit()``` function where further iterations usually decrease the Loglikelihood). 
Additionally every fitted object has now a new data variable ```em_traces``` that stores the value of the loglikelihood at each iteration. This can then be used to check convergence of the EM.

__Resulting behavior__: 
Computing time of most functions remain the same for larger datasets as those would exceed the minimum number of iterations anyway. For this reason the results also stay the same.
For smaller datasets however, several method including the ```fit()``` function and the ```sliding_event()``` are now notoriously slower (but better) because each tested starting point is susceptible to actually converge, hence to take longer.

PR #66, opened 2023-06-27
Improving em estimation

GWeindel:
Solves issues #61 #62 #63

### DIFF
--- a/src/hsmm_mvpy/models.py
+++ b/src/hsmm_mvpy/models.py
@@ -33,16 +33,17 @@ class hmp:
         location : float
             Minimum stage duration in milliseconds. 
         '''
-        if distribution == 'gamma':
-            from scipy.stats import gamma as sp_dist
-        elif distribution == 'lognormal':
-            from scipy.stats import lognorm as sp_dist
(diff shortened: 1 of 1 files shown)


src/hsmm_mvpy/models.py
```diff
@@ -89,6 +90,12 @@ def __init__(self, data, eeg_data=None, sfreq=None, offset=0, cpus=1, event_widt
         else: self.template = template
         self.events = self.cross_correlation(data.data.T)#adds event morphology
         self.max_d = self.durations.max()
+        self.em_method = em_method
+        if self.em_method == "mean":
```
jelmerborst, line 94: shouldn't there be an elif for "max" as well?

jelmerborst (approved):
great updates

src/hsmm_mvpy/models.py
GWeindel: No because the max method doesn't need this new data representation

GWeindel: As a result, using max method is also lighter in terms of memory

PR #70, opened 2023-06-30
Updated plot_topo_timecourse

jelmerborst:
- made sure topos cannot touch axes lines
- figure size is now dependent on `n_iter` being 1 or more
- added parameter `topo_size_scaling`: if `True` the width of the topologies scales width the displayed time interval. Ie, if 1000 ms are plotted and event_width = 50 ms, it's very small (this was the previous behavior). If `False`, topologies are fixed percentage of figure width. In both cases, `magnify` changes the topo size.


GWeindel (approved):
Looks perfect !

Issue #38, opened 2023-05-25
Random generation of starting points yields RuntimeWarning because of starting points larger than the mean RT


GWeindel:
Title explicit. For now, it is dealt with ignoring the warning as the error is not too bad. This however probably points to the fact that random generation of starting points is yet inefficient.

Ways to reproduce 
```python
init = hmp.models.hmp(hmp_data, sfreq=eeg_data.sfreq, bump_width=50, cpus=cpus)#Initialization of the model
selected = init.fit_single(number_of_sources-1, method='random', starting_points=10)#function to fit an instance of a 4 bumps model
```

GWeindel:
```python
                    proposal_p[1:-1,1] -= self.location
```
Fixes that error but generated parameters can then be negative. Future correction needs to integrate this constrain in the generation of random starting points

GWeindel:
Issue is actually still present. For now the error message is hidden:
```filterwarnings("ignore", category=RuntimeWarning)#Error in the generation of random see GH issue #38```

jelmerborst:
fixed by https://github.com/GWeindel/hsmm_mvpy/pull/60/commits/80210a72dceb8159b92ea91b7d7b57f498602316

jelmerborst:
apparently it is back, at least when calling the fit function:
![Screenshot 2023-06-27 at 17 02 24](https://github.com/GWeindel/hsmm_mvpy/assets/19282335/cd158a08-0a07-4f95-b682-1e9acde3cbdc)

jelmerborst:
fixed previously by:
`eventprobs[eventprobs < 0] = 0 #floating point precision error`
Not sure how this can happen anymore.

GWeindel:
This is very likely specific to the starting point generation of the fit function, I am currently on it

GWeindel:
Isn't an issue anymore since PR #71

PR #71, opened 2023-07-03
tutorial3 and models.py

jelmerborst:
- made some updates to tutorial 3

- changed `gen_random_stages`, as the previous version caused longer and longer randomly determined trial durations with more events (every gamma had a random duration between 0 and `mean_d`, the more stages that were estimated, the higher the sum of this number). The new version sets random moments for events between 0 and `mean_d`, and then calculates the corresponding gamma parameters. It takes into account minimum stage duration, and make sure that stages are at least 2 samples in length in addition to that.

- finally, I attempted to fix our ongoing warning (#38), as backward estimation throws an error if I used random starting points. I think that is due to #38 again, which is why I first updated `gen_random_stages`, which indeed reduced the number of times the warning occured, but didn't eliminate it. Second, I tried to catch our division-by-0 issue, see commented-out lines 404-408 in models.py (the np.clip in 402 avoids -0 values, whatever these are). However, this solution still gives me errors, so I suspect we should set `lkh` and `eventprobs` in a different way (among others: Convergence failed: one stage has been found to be null). I can try around further, but maybe @GWeindel knows directly what we should set those variables to.

I do think the current version is safe to merge.


GWeindel (approved):
Thanks, looks good, out of curiosity what is the advantage of using clip?

PR #72, opened 2023-07-04
check for div/0 in models.py

jelmerborst:
I checked what the reason for division by 0 is in estim_probs. It seems to be entirely due to extremely low probs, being represented as 0. This only happens when the magnitudes are still all 0. By avoiding division by 0 for the likelihood and eventprobs in those cases, the algorithm can just continue and estimate magnitudes. 

This also fixes backward_estimation with random starting points.


GWeindel (approved):
Isn't this way going to bias the likelihood for early iterations of the EM? I'll merge but we might want to keep that in mind

jelmerborst:
I agree that we should keep it in mind, but I don't think it does in the sense that those iterations are just extremely unlikely, as some trials do not fit at all.

PR #73, opened 2023-07-11
Major update towards 0.1.0

GWeindel:
Major changes:

- Removed the minimum number of iterations as this would force the EM to continue despite bad starting points, lead to dubious results
- The previous ```threshold``` argument has been replaced by two arguments: ```maximization``` and ```tolerance```: The former controls whether the maximization phase is done (bool) the later controls the criteria at which the EM is stopped in terms of log likelihood (defaults to ```1e-4``` as seen in other HMM applications). 
- Related to the previous point, the EM has now a ```max_iteration``` parameter to control how much iteration to do before stopping if the previous stopping rule was not met
- The fitted xarray object now contains the traces of the EM algorithm (```fit.traces```) also to inspect how the fit went
- Added the possibility of estimating several models based on several provided parameters or magnitudes. If one of them is 3D (e.g. in the case of magnitudes ```iteration X events X channel```), the ```single_fit``` function estimates one model per iteration. An additional parameter ```return_max``` controls whether only the highest likelihood model should be returned (True) or whether all should be returned (False). In the latter case the fitted object also contains an ```iteration``` dimension
- ```compute_max_events()``` is now computed based on the provided location not on the expected event duration. This increases the number of events that can be estimated when calling the default ```backward_estimation```
- added a function to generate magnitudes randomly or through grid search
- sliding_event() and fit() functions are still WIP

Minor:
- Moved import pandas as pd outside of the ```single_fit``` function. Also replaced it with a call to the specific function needed
- cpus is passed to ```fit_single``` to account for the fact that fit_single can also be paralellized f
- Changed the ordering of parameters to ```EM``` and ```single_fit```
- corrected index on the max EM method
- adapted location parameter based on simulations
- Added the possibility to pass an average position for the event to ```scale_parameters()``` to estimate parameters from average positions, not in use yet


src/hsmm_mvpy/models.py
```diff
@@ -191,101 +204,139 @@ def fit_single(self, n_events=None, magnitudes=None, parameters=None, threshold=
...
+            if len(np.shape(initial_m)) == 2:
+                parameters = [initial_p]
+                magnitudes = [initial_m]
+                if method == 'random':
+                    for _ in np.arange(starting_points):
+                        proposal_p = self.gen_random_stages(n_events, self.mean_d)
+                        proposal_p[parameters_to_fix] = initial_p[parameters_to_fix]
+                        #proposal_m = self._gen_mags(n_events, starting_points, method='random')
+                        # proposal_m[magnitudes_to_fix] = initial_m[magnitudes_to_fix]
+                        # magnitudes.append(proposal_m)
```
jelmerborst, line 218: remove commented lines

src/hsmm_mvpy/models.py
```diff
@@ -191,101 +204,139 @@ def fit_single(self, n_events=None, magnitudes=None, parameters=None, threshold=
...
+                        parameters.append(proposal_p)
+                    magnitudes = self._gen_mags(n_events, starting_points, method='random')
+                    magnitudes[:,magnitudes_to_fix,:] = np.tile(initial_m[magnitudes_to_fix], (len(magnitudes), 1, 1))
+
+                elif method == 'grid':
```
jelmerborst, line 223: didn't you want to remove grid?

src/hsmm_mvpy/models.py
```diff
@@ -191,101 +204,139 @@ def fit_single(self, n_events=None, magnitudes=None, parameters=None, threshold=
...
+                    parameters = self._grid_search(n_events+1, iter_limit=starting_points, method='grid')
+                    magnitudes = np.zeros((len(parameters), n_events, self.n_dims), dtype=np.float64)#Grid search is not yet done for mags
+                else:
+                    raise ValueError('Unknown starting point method requested, use "random" or "grid"')
+            elif len(np.shape(initial_m)) == 3:
```
jelmerborst, line 228: nice addition!

src/hsmm_mvpy/models.py
GWeindel: Only for the tutorials. The method will at some point include a grid search on the magnitudes but I don't have time to do it for now

src/hsmm_mvpy/models.py
GWeindel: Love it too :) So handy

src/hsmm_mvpy/models.py
```diff
@@ -191,101 +204,139 @@ def fit_single(self, n_events=None, magnitudes=None, parameters=None, threshold=
...
             eventprobs_sp = [x[3] for x in estimates]
             traces_sp = [x[4] for x in estimates]
-            max_lkhs = np.where(lkhs_sp == np.max(lkhs_sp))[0][0]
-            lkh = lkhs_sp[max_lkhs]
-            mags = mags_sp[max_lkhs]
-            pars = pars_sp[max_lkhs]
-            eventprobs = eventprobs_sp[max_lkhs]
-            traces = traces_sp[max_lkhs]
+            if return_max:
+                max_lkhs = np.where(lkhs_sp == np.max(lkhs_sp))[0][0]
```
jelmerborst, line 252: np.argmax?

src/hsmm_mvpy/models.py
```diff
@@ -298,16 +349,16 @@ def EM(self, n_events, magnitudes, parameters,  threshold, magnitudes_to_fix=Non
...
-        if threshold == 0 or n_events==0:
+        if not maximization or n_events==0:
             lkh_prev = lkh
             magnitudes_prev = initial_magnitudes
             parameters_prev = initial_parameters
             eventprobs_prev = eventprobs
         else:
             while i < max_iteration :#Expectation-Maximization algorithm
-                if i > min_iteration and lkh - lkh_prev < threshold:
+                if i > min_iteration and tolerance > lkh - lkh_prev:
```
jelmerborst, line 359: any reason for keeping min_iteration?

src/hsmm_mvpy/models.py
```diff
@@ -319,7 +370,7 @@ def EM(self, n_events, magnitudes, parameters,  threshold, magnitudes_to_fix=Non
                         #Average channel activity at those points
                         event_values = np.zeros((self.n_trials, self.n_dims))
                         for trial in range(self.n_trials):
-                            time = self.starts[trial]+ np.argmax(eventprobs[:, trial, event])
+                            time = self.starts[trial]+ np.argmax(eventprobs[:, trial, event])+1
```
jelmerborst, line 373: why +1? location?

src/hsmm_mvpy/models.py
```diff
@@ -464,16 +540,17 @@ def scale_parameters(self, eventprobs, n_events):
         params : ndarray
             shape and scale for the gamma distributions
         '''
-        averagepos = np.concatenate([np.arange(1,self.max_d+1)@eventprobs.mean(axis=1),
-                                     [self.mean_d]])#Durations
+        if eventprobs is not None:
+            averagepos = np.concatenate([np.arange(1,self.max_d+1)@eventprobs.mean(axis=1),
+                                         [self.mean_d]])#Durations
```
jelmerborst, line 545: I don't understand how you can take the mean of eventprobs to determine time. This will take the average of the probabilities. Shouldn't you do the dot product of samples and eventprobs?

src/hsmm_mvpy/models.py
```diff
@@ -534,7 +610,7 @@ def compute_max_events(self):
         '''
         Compute the maximum possible number of events given event width and mean or minimum reaction time
         '''
-        return int(np.min(self.durations)//(self.event_width_samples))
+        return int(np.rint(np.min(self.durations)//(self.location)))-2
```
jelmerborst, line 613: -1? Only the first stage doesn't have a location

src/hsmm_mvpy/models.py
GWeindel: Not really, do you think we should remove?

src/hsmm_mvpy/models.py
GWeindel: That is a leftover from previous trials, shouldn't be there I will remove

src/hsmm_mvpy/models.py
GWeindel: What I understand from what you suggest is what is being done in this line. First, take the mean of the probabilities by averaging across trials and then take the dotproduct with samples. Or do you mean to first do the dotproduct and then average?

jelmerborst (approved):
Check the -1/-2's, otherwise seems good to go.

src/hsmm_mvpy/models.py
jelmerborst: yes, it's only confusing and takes some time to check.

src/hsmm_mvpy/models.py
GWeindel: Indeed

src/hsmm_mvpy/models.py
jelmerborst: ah now I understand what you do. You can just directly do the dot product, it will directly take the average, no need to do that first.

src/hsmm_mvpy/models.py
GWeindel: Ah now I remember why I left it. We need a first run to verify ```lkh - lkh_prev```, in the previous versions we made an extra loop to get the first ```lkh_prev```, now it's replaced by enforcing at least 1 iteration. I suggest we keep as I'll need to track where it is declared

jelmerborst: what about setting lkh_prev to -inf instead to force the first run?

GWeindel: True that's enough, I'll make the change at some point

PR #74, opened 2023-07-12
added clustering for sliding event

jelmerborst:


src/hsmm_mvpy/models.py
```diff
@@ -868,26 +924,34 @@ def sliding_event(self, n_events=None, color='k', figsize=(12,3), verbose=True,
...
+        if cpus > 1:
+            if np.shape(magnitudes) == 2:
+                magnitudes = np.tile(magnitudes, (len(parameters), 1, 1))
+            with mp.Pool(processes=self.cpus) as pool:
+                estimates = pool.starmap(self.EM, 
+                    zip(itertools.repeat(1), magnitudes, parameters, 
+                        itertools.repeat(maximization), itertools.repeat(magnitudes_to_fix), 
+                        itertools.repeat(parameters_to_fix), itertools.repeat(max_iteration),
+                        itertools.repeat(tolerance)))
+        else:
```
GWeindel, line 942: Somehow your commits contains an old code. This if/else function is no longer necessary, we can just keep the content of the if loop and remove the if

src/hsmm_mvpy/utils.py
GWeindel, line 1: I think we need a separate module for the clustering as it seems relatively independent of the other functions in utils

GWeindel (commented):
Everything looks good except the conflict on estimate_single_event(). For the addition in utils, I wonder whether we shouldn't create another module to avoid cluttering the utils module. As the code you added seems independent of the rest of the utils methods I think it is a better thing to do

src/hsmm_mvpy/models.py
jelmerborst: Without the if/else it doesn't run because you get nested pools.

src/hsmm_mvpy/utils.py
jelmerborst: Sure, I'll move it to a new file.

src/hsmm_mvpy/models.py
GWeindel: Indeed, I forgot about that

jelmerborst:
Moved clustering to a separate module,  can be merged.

GWeindel (approved):

PR #75, opened 2023-07-12
checked tutorial 4, made some updates to visu

jelmerborst:
Just some small updates to tutorial 4, works again with current code base.

In addition, edited `plot_latencies_average` to calculate error bars correctly and build stacked barplot.


GWeindel (approved):
Looks good, thanks!

PR #78, opened 2023-07-19
clustering fixes, do not fix pars, grid decimation

jelmerborst:
Fixed two problems that caused clustering, or rather estimation, to go wrong:
- the decimation of the grid didn't work correctly, with a decimation of 2 there was no change to the number of models (with 5 PCs, 243 models)
- due to the addition of fix_pars to the sliding_event arguments, the pool inputs from sliding_event_mags were wrong, setting this new argument to 1... 

Can you somehow prevent the changes to the tutorials? I did not intend to change them (though I guess it's also not a problem).


GWeindel:
I cannot uncheck the changes to the tutorials, could you do a new PR? There are also ways for you to remove these files but I'll probably be longer

GWeindel:
Not entirely sure either, I think this will delete the files entirely

jelmerborst:
Do not merge, can be removed.

PR #80, opened 2023-07-26
fitting HMP models across conditions, specified using maps

jelmerborst:
See attached pdf for overview of options, and code for how this works with tutorial 4 (had to change extension to txt for some reason)

. Note that `mags_cl3` / `pars_cl3` are based on clustering, but they are just providing good starting points, can as well (or maybe better) be based on `fit`. I now provided starting points across conditions, but can also be done per condition.

[conditionsHMP.pdf](https://github.com/GWeindel/hsmm_mvpy/files/12173479/conditionsHMP.pdf)
[condition_effects.txt](https://github.com/GWeindel/hsmm_mvpy/files/12173493/condition_effects.txt)


GWeindel (approved):
As all of these are specific functions, the PR will not break any previous workflow.

Issue #81, opened 2023-07-26
Further work on condition specific codes


GWeindel:
The new PR #80 introduces a new way of fitting HMP across conditions while sharing (or not) some parameters.

Work left before definitive integration to a new version stored on pip:
- Merge EM_cond(), fit_single_cond() and estim_probs_cond() with general function (or other way around)
- assert missing stages/bumps
- add LOOCV
- update docstrings
- add barplots / lineplots
- check if random starting points work

jelmerborst:
`EM` and `estim_probs` have been merged. 

`fit_single` is harder to merge, as it can handle so many different inputs. Better approach is probably to add functions for some of the parts that are shared by `fit_single` and `fit_single_cond`

Another todo is to merge and/or simplify the plotting code.

GWeindel:
I ran into this error while executing tutorial 3:

```python
---------------------------------------------------------------------------
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/gweindel/anaconda3/envs/hmp/lib/python3.11/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "/home/gweindel/anaconda3/envs/hmp/lib/python3.11/multiprocessing/pool.py", line 51, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gweindel/owncloud/projects/RUGUU/hsmm-mvpy/src/hsmm_mvpy/utils.py", line 644, in LOOCV
    likelihood = model_left_out.estim_probs(fit.magnitudes.dropna('event').values, fit.parameters, n_events,True)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gweindel/owncloud/projects/RUGUU/hsmm-mvpy/src/hsmm_mvpy/models.py", line 807, in estim_probs
    if len(subset_epochs) == self.n_trials: #boolean indices
       ^^^^^^^^^^^^^^^^^^
TypeError: object of type 'bool' has no len()
"""
```

GWeindel:
[...]

PR #82, opened 2023-07-27
generalized EM/estim_probs for handling normal models and condition-b…

jelmerborst:
- as per #81 generalized EM and estim_probs to handle both single models and condition-based models
- removed n_events from EM as not necessary.


GWeindel (approved):
Looks great, I'll merge to main but given the complexity and length of the changes, we cannot exclude that something might be wrong. 
Once properly tested we'll integrate this and update the pip version to 0.1.1

PR #86, opened 2023-09-01
commented plotting code, speeded up compute_topologies

jelmerborst:
10x plotting speed-up (by speeding up compute_topologies: xarray is much cooler than we thought)

### DIFF
--- a/src/hsmm_mvpy/visu.py
+++ b/src/hsmm_mvpy/visu.py
@@ -86,38 +86,62 @@ def plot_topo_timecourse(channels, estimated, channel_position, init, time_step=
     from mne.viz import plot_brain_colorbar, plot_topomap
     from mne.io.meas_info import Info
     from mpl_toolkits.axes_grid1.inset_locator import inset_axes
-    from mpl_toolkits.axes_grid1 import make_axes_locatable
+    #from mpl_toolkits.axes_grid1 import make_axes_locatable
 
     return_ax = True
+    #if not times specified, plot average RT
     if times_to_display is None:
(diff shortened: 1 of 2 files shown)


src/hsmm_mvpy/visu.py
```diff
@@ -86,78 +86,94 @@ def plot_topo_timecourse(channels, estimated, channel_position, init, time_step=
...
-            channels = init.compute_topologies(channels, estimated, init.event_width_samples).data
-        times = init.compute_times(init, estimated, mean=True).data
-        channels[times == 0] = np.nan#removes the empty topologies, e.g. in the case of mutliple varying number of events
-
-    else:#assumes times already computed
+        if ydim is None and 'n_events' in estimated.dims: #and there are multiple different fits (eg backward estimation)
+            if estimated.n_events.count() > 1: #and we really have more than a singel fit
+                ydim = 'n_events' #set ydim to 'n_events'
+        if not skip_channels_computation:
+            channels = init.compute_topologies(channels, estimated, init.event_width_samples, ydim).data #compute topologies
```
GWeindel, line 115: We lost a condition here right? Doesn't it impact single fit plotting?

jelmerborst: This also works because ydim will be `None`, which evaluates to False apparently:
Screenshot 2023-09-01 at 14 08 50

GWeindel: OK then it's all good

GWeindel (approved):

PR #92, opened 2023-09-06
Improved specification of conditions to analyze

jelmerborst:
Instead of specifying to-be-analyzed conditions with a list of digits indicating the trials, one can now simply use the metadata, e.g., {'cue': ['SP','AC']}


GWeindel (approved):
Everything looks good, we just need some careful tests before shipping alll condition code with the next pip version

PR #93, opened 2023-09-06
assertion of missing stages, explanation with missing stages, improve…

jelmerborst:
Just more testing + an explanation if missing stages are indicated in condition specification. And reverse plotting order for condition fits, so it matches the specification.


GWeindel (approved):
Looks good :)

PR #94, opened 2023-09-08
rewrite plot_latencies, update compute_times

jelmerborst:
- rewritten plot_latencies to really handle bar and line plots. I propose to remove plot_latencies_average.
- added the ability to use hmp estimates as input to plot_latencies. If agreed, I'd argue for just removing the times-as-array option to clean up the code.
- added the ability to plot condition fits automatically with plot_latencies
- extended compute_times to be able to deal with missing stages when calculating durations
- extended compute_times to return time in ms instead of samples, also made this the default for plot_latencies. However, if time_step is provided, it takes precedence.

Still missing:
- error bars when using an hmp object (error calculation will be added to compute times)

### DIFF
--- a/src/hsmm_mvpy/models.py
+++ b/src/hsmm_mvpy/models.py
@@ -1196,7 +1196,7 @@ class hmp:
             return onsets
 
     @staticmethod        
-    def compute_times(init, estimates, duration=False, fill_value=None, mean=False, cumulative=False, add_rt=False, extra_dim=None):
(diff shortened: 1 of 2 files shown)


jelmerborst:
Figure_3
Figure_2

GWeindel (approved):
I absolutely agree to remove plot_latencies_average()
For the time as array however I do feel that the clarity won on the codebase is not enough compared to the flexibility this adds. So I would argue to keep it except if this really does make a big dfference

Great PR!!

PR #102, opened 2023-10-05
Loocv updates

jelmerborst:
Loocv:
- loocv consolidation to handle backward models, single models and condition based models
- new loocv_func that does loocv correctly: it applies a provided func to the data of n-1 subjects, and then calculates the likelihood on the left out subject. This function can be as complex as you want and also test multiple models at once. It's also not much slower in the end.

Cross collleration:
- fixed using numpy crosscorr method. Works now independent of template width and shape

In addition, while doing these a bunch of small fixes which only happen in edge cases, in particular:
- remove nans on pmfs
- min_iteration >= 1 instead of >
- checking for -inf

### DIFF
--- a/src/hsmm_mvpy/simulations.py
+++ b/src/hsmm_mvpy/simulations.py
@@ -60,7 +60,7 @@ def event_shape(event_width, event_width_samples, steps):
     template = template/np.sum(template**2)#Weight normalized
     return template
 
-def simulate(sources, n_trials, n_jobs, file, data_type='eeg', n_subj=1, path='./', overwrite=False, verbose=False, noise=True, times=None, seed=None, sfreq=None): 
+def simulate(sources, n_trials, n_jobs, file, data_type='eeg', n_subj=1, path='./', overwrite=False, verbose=False, noise=True, times=None, seed=None, sfreq=100): 
     '''
     Simulates n_trials of EEG and/or MEG using MNE's tools based on the specified sources
     
@@ -173,7 +173,7 @@ def simulate(sources, n_trials, n_jobs, file, data_type='eeg', n_subj=1, path='.
             # stim_onset occurs every x samples.
             events = np.zeros((n_trials, 3), int)
             stim_onsets =  2000 + max_trial_length * np.arange(n_trials) / (tstep*1000) #2000 = offset of first stim / in samples!
-            events[:,0] = stim_onsets#last event #maybe round?
+            events[:,0] = stim_onsets#last event 
             events[:,2] = 1#trigger 1 = stimulus 
 
             #Fake source, actually stimulus onset
--- a/src/hsmm_mvpy/utils.py
+++ b/src/hsmm_mvpy/utils.py
@@ -665,7 +665,7 @@ def transform_data(data, participants_variable="participant", apply_standard=Tru
     return data
     
 
-def LOOCV(data, init, participant, initial_fit, cpus=None, verbose=False):
+def loocv_calcs(data, init, participant, initial_fit, cpus=None, verbose=False):
     '''
     Fits model based on init settings and initial_fit parameters to data of 
     n - 1 (participant) participants, and calculates likelihood on the left-out
@@ -812,10 +812,10 @@ def loocv(init, data, estimate, cpus=1, verbose=True):
             loocv = []
             if cpus == 1: #not mp            
                 for participant in participants_idx:
-                    loocv.append(LOOCV(data, init, participant, model, verbose=verbose))
+                    loocv.append(loocv_calcs(data, init, participant, model, verbose=verbose))
             else: #mp
                 with mp.Pool(processes=cpus) as pool:
-                    loocv = pool.starmap(LOOCV,
+                    loocv = pool.starmap(loocv_calcs,
                                         zip(itertools.repeat(data), itertools.repeat(init),participants_idx,
                                             itertools.repeat(model),itertools.repeat(1),itertools.repeat(verbose)))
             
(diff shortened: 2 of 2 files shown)


src/hsmm_mvpy/models.py
```diff
@@ -807,7 +807,7 @@ def EM(self, magnitudes, parameters, maximization=True, magnitudes_to_fix=None,
         else:
             lkh_prev = -np.inf
             while i < max_iteration :#Expectation-Maximization algorithm
-                if i > min_iteration and tolerance > (lkh-lkh_prev)/np.abs(lkh_prev):
+                if i >= min_iteration and (np.isneginf(lkh) or tolerance > (lkh-lkh_prev)/np.abs(lkh_prev)):
```
GWeindel, line 810: Why the np.isneginf(lkh)? I didn't have this situation before I think

src/hsmm_mvpy/models.py
```diff
@@ -966,7 +966,7 @@ def estim_probs(self, magnitudes, parameters, n_events=None, subset_epochs=None,
             backward = np.zeros((self.max_d, n_trials, n_events), dtype=np.float64)
             # Computing forward and backward helper variable
             #  when stage = 0:
-            forward[:,:,0] = np.tile(pmf[:,0][np.newaxis].T,\
+            forward[:,:,0] = np.tile(np.concatenate([[0],pmf[:-1,0]])[np.newaxis].T,\
```
GWeindel, line 969: We agreed on this but somehow this is a strange fix, I'll investigate after the merge

src/hsmm_mvpy/models.py
```diff
@@ -1090,26 +1090,24 @@ def distribution_pmf(self, shape, scale, location):
...
             probabilty mass function for the distribution with given scale
         '''
         if scale == 0:
             warn('Convergence failed: one stage has been found to be null')
-        p = self.cdf(np.arange(1, self.max_d+1), shape, scale=scale, loc=location)
-        #Location is at least = to 1 given that range starts at 0, desirable as these are durations
-        p = np.diff(p, prepend=0)#going to pmf
+        p = self.pdf(np.arange(self.max_d)+1, shape, scale=scale)
+        p[:location] = 0
+        p[np.isnan(p)] = 0 #remove potential nans
```
GWeindel, line 1103: I never checked, does this happen regularly?

src/hsmm_mvpy/models.py
```diff
@@ -1457,7 +1455,13 @@ def compute_topologies(channels, estimated, init, extra_dim=None, mean=True, mea
             else:
                 times = np.round(xr.dot(estimated, estimated.samples, dims='samples'))
         
-            event_values = channels.sel(samples=times + shift).values
+            event_values = np.zeros((n_channels,n_trials,n_events))
```
GWeindel, line 1458: Awesome :)

src/hsmm_mvpy/simulations.py
```diff
@@ -60,8 +60,7 @@ def event_shape(event_width, event_width_samples, steps):
     template = template/np.sum(template**2)#Weight normalized
     return template
 
-
-def simulate(sources, n_trials, n_jobs, file, data_type='eeg', n_subj=1, path='./', overwrite=False, verbose=False, noise=True, times=None, location=1, seed=None): 
+def simulate(sources, n_trials, n_jobs, file, data_type='eeg', n_subj=1, path='./', overwrite=False, verbose=False, noise=True, times=None, seed=None, sfreq=None):
```
GWeindel, line 63: On the sfreq it probably needs a default or a catch for is None in line 135

GWeindel: I would say we can just go for a default at 100Hz

src/hsmm_mvpy/simulations.py
```diff
@@ -170,8 +172,8 @@ def simulate(sources, n_trials, n_jobs, file, data_type='eeg', n_subj=1, path='.
             sources_subj = sources[subj]
             # stim_onset occurs every x samples.
             events = np.zeros((n_trials, 3), int)
-            stim_onsets =  2000+max_trial_length * np.arange(n_trials)#2000 = offset of first stim
-            events[:,0] = stim_onsets#last event 
+            stim_onsets =  2000 + max_trial_length * np.arange(n_trials) / (tstep*1000) #2000 = offset of first stim / in samples!
+            events[:,0] = stim_onsets#last event #maybe round?
```
GWeindel, line 176: Yes rounding here won't hurt

src/hsmm_mvpy/utils.py
```diff
@@ -665,182 +665,575 @@ def transform_data(data, participants_variable="participant", apply_standard=Tru
     return data
     
 
-def LOOCV(data, participant, n_events, initial_fit, sfreq, event_width=50):
+def LOOCV(data, init, participant, initial_fit, cpus=None, verbose=False):
```
GWeindel, line 668: we should probably lower-case this function for consistency, I shouldn't have created it in uppercase...

src/hsmm_mvpy/utils.py
GWeindel, line 1: Everything looks absolutely great! We might have to keep an eye as the change in the code base are quite numerous but we'll see this on the fly after the merge

GWeindel (changes requested):
That is a great PR! I attached some reviews but nothing major, I think we can merge to main very soon

src/hsmm_mvpy/models.py
jelmerborst: This only happened in a very specific situation (feeding in already optimal mags/params, still going through EM), but I don't think it hurts. It might not be necessary with the i >= min_iteration, but I wasn't entirely sure.

src/hsmm_mvpy/models.py
jelmerborst: No, again only in an edge case where it tries to define super narrow pdfs. This fixes it, but then makes the likelihood go to -inf (otherwise you get errors)

src/hsmm_mvpy/simulations.py
jelmerborst: good point, yes 100hz would be good here!

src/hsmm_mvpy/simulations.py
jelmerborst: not necessary after all.

src/hsmm_mvpy/utils.py
jelmerborst: then it needs a different name, I will fix it.

src/hsmm_mvpy/utils.py
jelmerborst: sounds like a plan. Let's merge after my next commit, then run the tutorials and prep the next workshop to see where things break.

src/hsmm_mvpy/models.py
GWeindel: Yes it can stay, I was just curious

src/hsmm_mvpy/models.py
GWeindel: We can leave it to avoid the errors but then make a more clear warning, but not necessary right now

GWeindel (approved):
All green!

PR #103, opened 2023-10-11
location fix in HMP for loocv

jelmerborst:
self.location was set incorrectly when provided, which introduced a problem with LOOCV where it was copied from an existing model. I do not think line 111 is used for anything else, but that should be checked.

### DIFF
--- a/src/hsmm_mvpy/models.py
+++ b/src/hsmm_mvpy/models.py
@@ -109,7 +109,7 @@ class hmp:
         if location is None:
             self.location = int(self.event_width / self.steps//2)+1
         else:
-            self.location = location
+            self.location = int(np.rint(location))
         durations = data.unstack().sel(component=0).rename({'epochs':'trials'})\
             .stack(trial_x_participant=['participant','trials']).dropna(dim="trial_x_participant",\
             how="all").groupby('trial_x_participant').count(dim="samples").cumsum().squeeze()


src/hsmm_mvpy/models.py
```diff
@@ -108,7 +108,8 @@ def __init__(self, data, epoch_data=None, sfreq=None, cpus=1, event_width=50, sh
         self.event_width_samples = int(np.round(self.event_width / self.steps))
         if location is None:
             self.location = int(self.event_width / self.steps//2)+1
-        else: self.location =  int(np.rint(location / self.steps))
+        else:
```
GWeindel, line 111: Indeed but the int(np.rint() insured that it was an int, perhaps then just add an assert?

GWeindel (approved):
Fine by me, we should just add the assert in case a location is specified in float

src/hsmm_mvpy/models.py
jelmerborst: sure, or leave the `int(np.rint())` and just remove `/ self.steps` (which made the location go to 0 and it estimated overlapping events).

Do you have a preference for` int(np.rint())` or assert?

GWeindel: We can just keep ```int(np.rint())``` I would say

PR #105, opened 2023-10-13
cross corr shifted to mid template

jelmerborst:
major change: shift the cross correlation to the mid of the template, including changes to switching times and topologies


GWeindel (approved):
Awesome!

PR #106, opened 2023-10-14
event_times reintroduced

jelmerborst:


GWeindel (approved):
I cannot view the diff in the notebooks but I assume everything is good and will take a closer look soon

PR #108, opened 2023-10-14
Bootstrapping on single_fit() instead of fit()

GWeindel:
- Replaced bootstrapping function on the fit function to bootstrapping on single_fit()

the bootstrapping is now performed on the fit object provided when the function is called. The fit on boostrapped data takes the same number of events as the fit object provided and uses the provided fits estimation of parameters and magnitudes as starting point

Also minor fixed on models.py


jelmerborst (approved):
looks great!

PR #118, opened 2023-10-25
Filtering crosscorr

jelmerborst:
- added filtering as an option to transform_data
- some small bug fixes

### DIFF
--- a/src/hsmm_mvpy/models.py
+++ b/src/hsmm_mvpy/models.py
@@ -122,9 +122,7 @@ class hmp:
             dur_dropped_na = durations
             starts = np.array([0])
             ends = np.array([dur_dropped_na.data-1])
-        self.starts = starts
-        self.ends = ends
-        self.durations = self.ends-self.starts+1
+        self.durations = ends-starts+1
         if durations.trial_x_participant.count() > 1:
             self.named_durations =  durations.dropna("trial_x_participant") - durations.dropna("trial_x_participant").shift(trial_x_participant=1, fill_value=0)
             self.coords = durations.reset_index('trial_x_participant').coords
@@ -139,7 +137,7 @@ class hmp:
             self.template = self._event_shape()
         else: self.template = template
         self.max_d = self.durations.max()
-        self.events = self.cross_correlation(data.data.T)#adds event morphology
+        self.events = self.cross_correlation(data.data.T, ends, starts)#adds event morphology
         self.em_method = em_method
         self.estimate_magnitudes = estimate_magnitudes
         self.estimate_parameters = estimate_parameters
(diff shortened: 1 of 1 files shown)


jelmerborst:
- added warning to filtering
- reshaped self.events to samp x trials x dim, both to remove self.data_matrix and make it in line with most other structures

tutorials/Workshop/dataset_raw_raw.fif
GWeindel, line 1: Is this needed?

tutorials/Workshop/dataset_raw_raw_generating_events.npy
GWeindel, line 1: Is this needed?

src/hsmm_mvpy/models.py
```diff
@@ -926,30 +919,27 @@ def estim_probs(self, magnitudes, parameters, n_events=None, subset_epochs=None,
...
             # computes the gains, i.e. congruence between the pattern shape
             # and the data given the magnitudes of the sensors
-            gains = gains + self.events[:,i][np.newaxis].T * magnitudes[:,i]-magnitudes[:,i]**2/2
+            for tr in range(n_trials):
+                gains[:, tr, :] =  gains[:, tr, :] + self.events[:,tr,i][np.newaxis].T * magnitudes[:,i]-magnitudes[:,i]**2/2
+        
         gains = np.exp(gains)
         probs = np.zeros([self.max_d, n_trials,n_events], dtype=np.float64) # prob per trial
         probs_b = np.zeros([self.max_d, n_trials,n_events], dtype=np.float64)# Sample and state reversed
         for trial in np.arange(n_trials):
```
GWeindel, line 936: This part can then be rewritten, I can take a look this afternoon

GWeindel (changes requested):
Awesome! are there any performance differences?

For the gains to prob in models.py we can now make things faster by just using gains and probably invert probs to probs_b easily. I can try this afternoon

tutorials/Workshop/dataset_raw_raw.fif
jelmerborst: No, don't know where this comes from.

tutorials/Workshop/dataset_raw_raw_generating_events.npy
jelmerborst: same

src/hsmm_mvpy/models.py
jelmerborst: Yeah of course, that makes sense. Even better :)

GWeindel:
After careful consideration we together decided to halt that development as it seems that moving from the non-sparse format of ```self.events``` (````n_samples X channels```) to a sparse format (```n_trials x max_n_samples_per_trial x channels```) yields performance issue (this branch is between 1/5th-1/10th slower than the current main). Later development might either find a way around that slow down so we close the PR but keep the branch.

PR #119, opened 2023-10-27
filtering in transform data

jelmerborst:
- added filtering option to transform_data
- small bug fixes

### DIFF
--- a/src/hsmm_mvpy/utils.py
+++ b/src/hsmm_mvpy/utils.py
@@ -1374,8 +1374,38 @@ def event_times(data, times, channel, stage):
 
     return brp_data    
     
-def condition_selection(hmp_data, epoch_data, condition_string, variable='event'):
-    return hmp_data.where(hmp_data[variable] == condition_string, drop=True)
+def condition_selection(hmp_data, epoch_data, condition_string, variable='event', method='equal'):
+    '''
+    condition_selection select a subset from hmp_data. It selects epochs for which
+    'condition_string' is in 'variable' based on 'method'.
+
+    Parameters
+    ----------
+    hmp_data : xr.Dataset
+        transformed EEG data for hmp, from utils.transform_data
+    epoch_data : deprecated
+    condition_string : str | num
+        condition indicator for selection
+    variable : str
+        variable present in hmp_data that is used for condition selection
+    method : str
+        'equal' selects equal trials, 'contains' selects trial in which conditions_string
+        appears in variable
+
+    Returns
+    -------
+    dat : xr.Dataset
+        Subset of hmp_data.
+        
+    '''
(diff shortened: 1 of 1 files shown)


src/hsmm_mvpy/models.py
```diff
@@ -188,8 +188,7 @@ def cross_correlation(self,data):
         events = np.zeros(data.shape)
         for trial in range(self.n_trials):#avoids confusion of gains between trials
             for dim in np.arange(self.n_dims):
-                dat = data[self.starts[trial]:self.ends[trial]+1, dim]
-                events[self.starts[trial]:self.ends[trial]+1,dim] = correlate(dat, self.template, mode='same', method='direct')
+                events[self.starts[trial]:self.ends[trial]+1,dim] = correlate(data[self.starts[trial]:self.ends[trial]+1, dim], self.template, mode='same', method='direct')
```
GWeindel, line 191: Isn't that a residual from PR #118 ?

src/hsmm_mvpy/utils.py
```diff
@@ -1344,9 +1375,8 @@ def event_times(data, times, channel, stage):
     return brp_data    
     
 def condition_selection(hmp_data, epoch_data, condition_string, variable='event'):
-    unstacked = hmp_data.unstack().where(epoch_data[variable].str.contains(condition_string),drop=True)
-    stacked = stack_data(unstacked)
-    return stacked
+    return hmp_data.where(hmp_data[variable] == condition_string, drop=True)
```
GWeindel, line 1378: Does this have a link with the recent weirdness in condition selection? Also contains is useful (e.g. Tutorial 0, raw data section), so we should rather have a parameter, e.g. method' that by default takes 'equal' but can also be adjusted to 'contains'

GWeindel (changes requested):
- Reimplement contains as parameter
- Check the correctness of the left-over from PR#118

src/hsmm_mvpy/models.py
jelmerborst: Yes but it doesn't change anything and is slightly more efficient.

src/hsmm_mvpy/utils.py
jelmerborst: No because for the condition code you actually do not split the data in conditions. I made this change because the original version crashed my code, and this seemed much simpler and do the same job. I'll add `contains` as an option.

src/hsmm_mvpy/models.py
GWeindel: Ah yes I didn't pay attention to the first removed line

jelmerborst:
done

GWeindel (approved):
With docstrings, awesome!

PR #121, opened 2023-11-14
condition selection keeping metadata

jelmerborst:
- changed condition selection so it maintains metadata and is independent from epoch_data. Works for my data and Tutorial 4.


GWeindel (approved):
Awesome!

PR #126, opened 2023-12-26
Condition selection 'contains' method bug

rickdott:
stacking/unstacking introduced nans into coordinate of 'object' type, meaning .str.contains() cannot match it. Fixed by replacing na values with "" (empty string), which should never match with condition_string, unless of course it is the empty string.


GWeindel (approved):
It's a good catch and fix, thanks!

PR #131, opened 2024-03-11
Turn off the location for the last round of EM 

GWeindel:
Turn off the location for the last round of EM : Event probabilities are therefore given without location to ensure that times are not influenced by the locations choosen or developed. Doesn't change event detection and EM solutions, just time estimations given by the event probabilities

For a reproducible example of the PR just run the README with very short stage duration, the benefit is apparent in all plots, e.g.:
```python 
scales = np.array([60, 25, 200, 100, 80])/shape #Mean duration of the time between each event in ms
names = simulations.available_sources()[[44,33, 22, 55,0]]#Which source to activate for each event (see atlas when calling simulations.available_sources())

sources = []
for source in zip(names, scales):#One source = one frequency/event width, one amplitude and a given by-trial variability distribution
    sources.append([source[0], frequency, amplitude, gamma(shape, scale=source[1])])

# Function used to generate the data
file = simulations.simulate(sources, n_trials, cpus, 'dataset_README',  overwrite=True, sfreq=sfreq, seed=1)
```
Leads to : 
Screenshot_20240311_102334

with the current branch vs. main : 
Screenshot_20240311_102247

### DIFF
--- a/hmp/utils.py
+++ b/hmp/utils.py
@@ -759,7 +759,7 @@ def transform_data(data, participants_variable="participant", apply_standard=Fal
                 if zscore_acrossPCs:
                     data = data.stack(trial=[participants_variable,'epochs']).groupby('trial').map(zscore_xarray).unstack()
                 else:
-                    data = data.stack(trial=[participants_variable,'epochs','component']).groupby('trial').map(zscore_xarray).unstack()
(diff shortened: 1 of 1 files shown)


jelmerborst (approved):
this looks great, also very useful to not have the LOOCV biased by locations

PR #132, opened 2024-03-12
Dev

GWeindel:
## Simplified fit function and removed deltas
I removed all the parameters that weren't used just to simplify, this PR removed
- decimate
- min_iteration
- grid_points
- start
Arguments from the fit() function as they don't seem useful, if needed in the future it can be found from here.

More importantly, based on simulations it looks like the delta parameter used up to now is only necessary for the detection of the first event (as there is a bias in the likelihood when we split the RTs in two). It doesn't seem necessary to have this counterfactual model for following events

## removed filtering of non-converged models 
as it seems addressed by recent progress in location definition and was leading to useless rejection of a good proposal by the fit function

## Switched default location to event_width_samples
Overall choosing this default seems the best, fastest, strategy, if shorter stage duration are expected then this can be decreased or location_correlation_threshold can be used

## Fixed problem with ultimate location not being 0
By forcing it to 0 in the propose_fit_params() function instead of the EM

## minor changes
in README, tutorials and simulation functions


jelmerborst (approved):
very nice clean-up, was long overdue :)

PR #134, opened 2024-04-02
gamma scaling in loocv

jelmerborst:
Update that scales the gamma parameters to the average duration of the left-out subject before calculating their likelihood. Slightly less conservative than previous version, can find evidence for at least one extra event.

Old version:
image

New version:
image

for a dataset with certainly at least two events:
image


GWeindel (approved):
Looks good, I think we need to re-check the tutorial now though

PR #136, opened 2024-04-04
simulations were not indexed

jelmerborst:


GWeindel (commented):
Simulations are left out on purpose as this should only be for specific cases (+ might force to download mne data for simulations although I didn't check). 
For resample it's not yet functional as I'm still working on it, do you need it now?

PR #137, opened 2024-04-25
fit fix + simulation fix

jelmerborst:


GWeindel (approved):
That all makes sense, time -> time_out in simulation is just convention?

jelmerborst:
neh, time_out gave issues when I wanted to generate multiple participants, I believe, or multiple datasets – something like that.

PR #140, opened 2024-05-16
Sped up centering operation in transform_data

rickdott:
Using (Boehm, 2014) data, participants 1-5:

Before:
CPU times: user 52.8 s, sys: 516 ms, total: 53.3 s
Wall time: 53.3 s

After:
CPU times: user 6.32 s, sys: 91.4 ms, total: 6.42 s
Wall time: 6.43 s

Should also scale better to larger datasets.

### DIFF
--- a/hmp/utils.py
+++ b/hmp/utils.py
@@ -505,9 +505,11 @@ def _center(data):
     '''
     zscore of the data
     '''
-    non_nan_mask = ~np.isnan(data.values)
-    if non_nan_mask.any(): #if not everything is nan, calc zscore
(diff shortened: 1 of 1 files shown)


GWeindel (approved):
Awesome, thanks !

rickdott:
Does now print this:
```
RuntimeWarning: Mean of empty slice
  mean_last_dim = np.nanmean(data.values, axis=-1)
```
When running, let me know if this is an issue then we can catch the warning so it does not print:
https://stackoverflow.com/questions/29688168/mean-nanmean-and-warning-mean-of-empty-slice

PR #148, opened 2024-10-16
Removed stacked epoch data from HMP model when not necessary, saves RAM

rickdott:
As discussed, saves a lot of memory when using multithreading.


GWeindel (approved):
Looks good, nice RAM economy :) :1st_place_medal:

PR #149, opened 2024-10-22
Included Testing notebook & Started writing the testing workflow for every push on main.

nmararo:


GWeindel:
Hi Mara, I think you're not doing the PR on the correct branch. Right now it says from main to publish-and-release but you should make a PR from your version to main

nmararo:
Strange, I tried doing it the other way around the first time around, but
it didn't allow me to do it. I will try it that way once more.

PR #172, opened 2025-01-21
fix easy linting errors visu.py

jelletreep:
Low hanging fruit, now back to 10 errors which are all: too many statements/branches/arguments

### DIFF
--- a/hmp/visu.py
+++ b/hmp/visu.py
@@ -32,7 +32,6 @@ def plot_topo_timecourse(
     title=False,
     ax=None,
     sensors=False,
-    skip_channels_computation=False,
     contours=6,
     event_lines="tab:orange",
     colorbar=True,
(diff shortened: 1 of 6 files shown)


hmp/visu.py
```diff
@@ -1045,7 +1059,9 @@ def plot_erp(
     times, data, color="k", ax=None, minmax_lines=None, upsample=1, bootstrap=None, label=None
 ):
     """
-    Plot the ERP based on the times extracted by HMP (or just stimulus and response and the data extracted from ```erp_data```.
+    Plot the ERP based on the times extracted by HMP.
+
+    (or just stimulus and response and the data extracted from ```erp_data```.
```
GWeindel, line 1064:
```suggestion
    Either around an event or just stimulus and response and the data extracted from ```erp_data```.
```

jelletreep:
@GWeindel, I committed your suggestion.
There is a conflict to solve, mainly `channel_data` <> `epoch_data` as far as i can see. It probably should be channel data as it is part of your bug fix on monday, but maybe good if you check and fix the conflict and then merge?

Issue #173, opened 2025-01-27
SKLearn creating copy of the data


GWeindel:
Currently when applying PCA the data is being copyied by sklearn, should not be the case, can be controlled by passing ```copy=False``` when calling ```PCA(n_components=n_comp, svd_solver="full") ```

PR #175, opened 2025-01-27
Memory management improvement

GWeindel:
\Removed unnecessary copies of heavy files, place warning in the function that can actually change the data inplace


GWeindel:
Fixes #173

qubixes (approved):
Seems like a sensible approach to save memory!

PR #176, opened 2025-01-29
Refactor models.py

qubixes:
The code doesn't work yet. At least the LOOCV integration is now broken and should be fixed.

### DIFF
--- a/hmp/models/backward.py
+++ b/hmp/models/backward.py
@@ -1,18 +1,13 @@
 """Models to estimate event probabilities."""
 
 import gc
-import itertools
-import multiprocessing as mp
-from itertools import cycle, product
-from warnings import resetwarnings, warn
 
-import matplotlib.pyplot as plt
 import numpy as np
-import xarray as xr
-from pandas import MultiIndex
 import pandas as pd
-from scipy.signal import correlate
-from scipy.stats import norm as norm_pval
+import xarray as xr
+
+from hmp.models.base import BaseModel
+from hmp.models.fixedn import FixedEventModel
 
 try:
     __IPYTHON__
@@ -21,8 +16,7 @@ except NameError:
     from tqdm import tqdm
 
 default_colors = ["cornflowerblue", "indianred", "orange", "darkblue", "darkgreen", "gold", "brown"]
-from hmp.models.base import BaseModel
-from hmp.models.fixedn import FixedEventModel
+
 
 class BackwardEstimationModel(BaseModel):
     def __init__(self, *args, max_events=None, min_events=0, max_starting_points=1,


hmp/models/base.py
```diff
@@ -0,0 +1,205 @@
...
+except NameError:
+    from tqdm import tqdm
+
+
+@dataclass
+class EventProperties():
+    sfreq: float
+    steps: float
+    shape: float
+    width: int
```
GWeindel, line 32: This can be floats I would assume

hmp/models/base.py
```diff
@@ -0,0 +1,205 @@
...
+    width_samples: int
+    location: int
+    template: Any
+
+    @classmethod
+    def from_standard_data(cls, data, sfreq=None, shape=2, width=50, template=None, location=None):
```
GWeindel, line 38: While the EventProperties class is a very good structure, I think this is a bit confusing 
```suggestion
    def expected_event(cls, data, sfreq=None, shape=2, width=50, template=None, location=None):
```

hmp/models/base.py
```diff
@@ -0,0 +1,205 @@
...
+    from tqdm.notebook import tqdm
+except NameError:
+    from tqdm import tqdm
+
+
+@dataclass
+class EventProperties():
+    sfreq: float
+    steps: float
+    shape: float
```
GWeindel, line 31: I understand why you made it so but in the end shape is an argument for the distribution. Do you think we can build the distribution here in the class instead of BaseModel?

hmp/models/base.py
```diff
@@ -0,0 +1,205 @@
...
+        self.distribution = distribution
+        self.trial_data = trial_data
+        self.events = event_properties
+        self.pdf = sp_dist.pdf
+
+    def compute_max_events(self):
+        """Compute the maximum possible number of events given event width minimum reaction time."""
+        return int(np.rint(np.percentile(self.durations, 10) // (self.location)))
+
+    def gen_random_stages(self, n_events):
```
GWeindel, line 152: Shouldn't this be in fixedn?

hmp/models/fixedn.py
```diff
@@ -317,8 +139,8 @@ def fit_n(
                 )
             else:
                 print(f"Estimating {n_events} events model")
-        if cpus is None:
-            cpus = self.cpus
+        # if cpus is None:
+            # cpus = self.cpus
```
GWeindel, line 143: Shouldn't this be simply removed?

hmp/models/forward.py
```diff
@@ -0,0 +1,295 @@
...
+try:
+    __IPYTHON__
+    from tqdm.notebook import tqdm
+except NameError:
+    from tqdm import tqdm
+
+default_colors = ["cornflowerblue", "indianred", "orange", "darkblue", "darkgreen", "gold", "brown"]
+
+
+class ForwardEstimationModel(BaseModel):
```
GWeindel, line 28: Cumulative is probably a better as forward can refer to the forward variable computed during expectation-maximization
```suggestion
class CumulativeEstimationModel(BaseModel):
```

hmp/models/forward.py
```diff
@@ -0,0 +1,295 @@
+"""Models to estimate event probabilities."""
+
+import gc
+import itertools
+import multiprocessing as mp
+from itertools import cycle, product
+from warnings import resetwarnings, warn
+
```
GWeindel, line 8: There's probably a bunch of librabies that we no longer need in the individual modules right?

hmp/trialdata.py
```diff
@@ -0,0 +1,118 @@
...
+    def max_duration(self):
+        return self.durations.max()
+
+    @cached_property
+    def mean_duration(self):
+        return self.durations.mean()
+
+    @cached_property
+    def durations(self):
+        return self.ends - self.starts + 1
```
GWeindel, line 89: Isn't this slowing things (slightly) down? That data shouldn't be too heavy so I would say we should compute it once and then just refer to that data (e.g. see the repeated calls in fixedn.estim_probs)

GWeindel: Same for the mean_durations and max_durations

hmp/models/base.py
GWeindel: Additionally, I think giving the data is a bit confusing as this is really only about the events we except in the timeseries. In fact we only need the sampling frequency. So rather than asking participant to provide the data we just ask for the sampling frequency.

GWeindel (changes requested):
I think this is a great new structure much more modular and nicely allowing future new model implementations. I have made some comments but those are mainly details and small re-organizations

hmp/models/base.py
qubixes: Looks like it indeed!

hmp/models/base.py
qubixes: I was planning to create a distribution class, but @maartenschermer might be busy with that?

hmp/models/base.py
qubixes: I have renamed it create_expected, does that make sense?

hmp/trialdata.py
qubixes: Cached property computes it once and caches it, so it will only be computed once (when it is first needed).

hmp/models/base.py
qubixes: What do you mean exactly?

GWeindel: Oups I meant the width should be floats

hmp/trialdata.py
GWeindel: Oh OK perfect

GWeindel:
Maximization in the model was a placeholder for the transform only method, given that we now have an actual transform method it can be removed. I can do it in the next days

hmp/utils.py
```diff
@@ -185,10 +186,10 @@ def read_mne_data(
         print(f"Processing participant {participant}'s {dict_datatype[epoched]} {pick_channels}")
 
         # loading data
-        if epoched is False:  # performs epoching on raw data
-            if ".fif" in participant:
+        if not epoched:  # performs epoching on raw data
```
GWeindel, line 189:
```suggestion
        if epoched is False:  # performs epoching on raw data
```

tests/old_test_integration.py
GWeindel, line 1: This file can be removed no?

GWeindel (changes requested):
Looks good to me, pending the two suggestions

hmp/utils.py
```diff
@@ -1112,6 +1009,7 @@ def event_topo(
     """
     if estimate_method is None:
         estimate_method = "max"
+<<<<<<< HEAD
```
GWeindel, line 1012: Conflict to deal with, 516db5c is correct

qubixes: Done

tests/old_test_integration.py
qubixes: The loocv tests are not there yet though. Shall we keep it until those are in there as well?

GWeindel (approved):

hmp/models/forward.py
qubixes: I removed most of the unused imports.

tests/old_test_integration.py
GWeindel: AH yes good point

PR #178, opened 2025-04-08
Modify cumulative model for new structure

qubixes:


tests/test_cumulative.py
```diff
@@ -54,15 +54,13 @@ def test_fixed_simple():
...
-    # fit the model
+    # Cumulative estimation
+    model = CumulativeEstimationModel(event_properties, step=25)
     model.fit(trial_data_b)
-    # Transform the data
     estimates = model.transform(trial_data_b)
 
     # testing if bacward identifies the 3 real events
-    # assert np.isclose(model.submodels[3].magnitudes, true_model.magnitudes, atol=1).all()
+    assert np.isclose(model.fitted_model.magnitudes, true_model.magnitudes, atol=1).all()
```
GWeindel, line 63: Do we really need the attribute `fitted_model`? For the cumulative method, given that there is no other model stored I would just make the same structure as `fixedn`

qubixes: Since we're using encapsulation (Cumulative model *has* a fixed model) and not inheritance (Cumulative model *is* a fixed model), we do need the attribute. What we could do is pass through the parameters and magnitudes with the `__getattribute__` method.

GWeindel (approved):
Looks all good, thanks!

PR #181, opened 2025-04-09
Moving parameters that are specific to the fit from the init to the fit method

GWeindel:
It makes more sense that:
        tolerance=1e-4,
        max_iteration=1e3,
        min_iteration=1,
        starting_points=1,
        max_scale=None,
are defined in the fit funciton rather than the init. thre reason is that a `fixed_n` model is always defined by n_event, but that fits can be called with different parameters to the estimation.

### DIFF
--- a/hmp/models/backward.py
+++ b/hmp/models/backward.py
@@ -19,9 +19,7 @@ default_colors = ["cornflowerblue", "indianred", "orange", "darkblue", "darkgree
 
 
 class BackwardEstimationModel(BaseModel):
-    def __init__(self, *args, max_events=None, min_events=0,  **kwargs):
-        self.max_events = max_events
-        self.min_events = min_events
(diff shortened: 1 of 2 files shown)


qubixes:
@GWeindel Thanks for thinking about this. I had also thought about this before, but I had obviously come to a different conclusion. I was trying to follow mostly in the way of sklearn, see for example https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#sklearn.linear_model.LogisticRegression 
Is there a strong opinion about doing it differently?

GWeindel:
Ah I see they do it the same way in MNE https://mne.tools/stable/generated/mne.preprocessing.ICA.html

GWeindel:
OK I guess it's best to align with those

PR #184, opened 2025-04-17
Refactored transform

maartenschermer:
Refactored the transform function, no big changes.
Includes a notebook comparing results of old and new method (which somehow are not equal, so some work to be done still)

### DIFF
--- a/tests/test_backward.py
+++ b/tests/test_backward.py
@@ -6,6 +6,7 @@ from hmp import simulations
 from hmp.models import BackwardEstimationModel, FixedEventModel
 from hmp.models.base import EventProperties
 from hmp.trialdata import TrialData
+from hmp import preprocessing
 
 
 from test_io import init_data
--- a/tests/test_cumulative.py
+++ b/tests/test_cumulative.py
@@ -6,7 +6,7 @@ from hmp import simulations
 from hmp.models import CumulativeEstimationModel, FixedEventModel
 from hmp.models.base import EventProperties
 from hmp.trialdata import TrialData
-from hmp.transform import DataTransformer
+from hmp import preprocessing
 
 from test_fixed import init_data
 
--- a/tests/test_fixed.py
+++ b/tests/test_fixed.py
@@ -7,6 +7,8 @@ from hmp import simulations
 from hmp.models import FixedEventModel
 from hmp.models.base import EventProperties
 from hmp.trialdata import TrialData
+from hmp import preprocessing
+
 
 from test_io import init_data
 
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -5,6 +5,7 @@ import numpy as np
 from hmp import io
 from hmp import simulations
 from hmp import utils
+from hmp import preprocessing
 from hmp.models.base import EventProperties
 from hmp.trialdata import TrialData
 from hmp.models import FixedEventModel
@@ -38,8 +39,8 @@ def init_data():
 
 def test_save_dat():
     event_b, event_a, epoch_data, positions, sfreq, n_events = init_data()
-    hmp_data = utils.transform_data(epoch_data, n_comp=2,)
-    data_b = utils.participant_selection(hmp_data, 'b')
+    hmp_data = preprocessing.Preprocessing(epoch_data, n_comp=2,)
+    data_b = utils.participant_selection(hmp_data.data, 'b')
     event_properties = EventProperties.create_expected(sfreq=data_b.sfreq)
     trial_data_b = TrialData.from_standard_data(data=data_b, template=event_properties.template)
 
--- a/tests/test_plots.py
+++ b/tests/test_plots.py
@@ -5,13 +5,15 @@ from hmp.models import FixedEventModel
 from hmp.models.base import EventProperties
 from hmp.trialdata import TrialData
 from hmp.visu import plot_topo_timecourse
+from hmp import preprocessing
+
 
 from test_io import init_data
 
 
 def test_plot():
     _, _, epoch_data, positions, sfreq, n_events = init_data()
-    hmp_data = hmp.utils.transform_data(epoch_data, n_comp=2,)
+    hmp_data = preprocessing.Preprocessing(epoch_data, n_comp=2,)
     # Testing one event less in one condition
     mags_map = np.array([[0, 0, -1],
                          [0, 0, 0]])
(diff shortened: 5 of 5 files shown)


qubixes (changes requested):
Just a quick review from my side. I haven't checked the validity, I'll leave that to @GWeindel. The code looks solid overall.

qubixes:
Btw, I don't see any removed lines. Is the old code still in there?

hmp/transform.py
```diff
@@ -14,18 +14,51 @@ class ApplyZScore(Enum):
...
+    DONT_APPLY = 'dont_apply'
 
     def __str__(self) -> str:
         return self.value
 
+    def apply(self) -> bool:
+        return self != ApplyZScore.DONT_APPLY
 
-class Method(Enum):
+    @staticmethod
```
qubixes, line 25: Doesn't a classmethod make more sense?

hmp/transform.py
```diff
@@ -14,18 +14,51 @@ class ApplyZScore(Enum):
...
+            return label
+        elif label is False or label is None:
+            return ApplyZScore.DONT_APPLY
+        elif label in ('trial', True):
+            return ApplyZScore.TRIAL
+        elif label == 'participant':
+            return ApplyZScore.PARTICIPANT
+        elif label == 'all':
+            return ApplyZScore.ALL
+        else:
```
qubixes, line 37: I think a KeyError makes more sense? Or perhaps a ValueError?

hmp/transform.py
```diff
@@ -150,21 +183,23 @@ def __init__(
...
             np.isnan(data.groupby("participant",
                                   squeeze=False).mean(["epochs", "samples"]).data.values)) != 0:
             raise ValueError("at least one participant has an empty channel")
 
-        if not isinstance(method, Method) and method is not None:
-            raise ValueError(f"method {method} is unknown, choose either {', '.join([e.value for e in Method])} or None")   # noqa: E501
+        try:
+            method = AnalysisMethod(method)
+        except NotImplementedError as e:
+            raise NotImplementedError(f"Unknown method: {method!r}; valid options: {', '.join([e.value for e in AnalysisMethod])} or None") from e  # noqa: E501
```
qubixes, line 200:
```suggestion
            raise NotImplementedError(f"Unknown analysis method: {method!r}; valid options: {', '.join([e.value for e in AnalysisMethod])} or None") from e  # noqa: E501
```

hmp/transform.py
```diff
@@ -150,21 +183,23 @@ def __init__(
...
-        if apply_zscore is True:
-            apply_zscore = ApplyZScore.TRIAL   # defaults to trial
-
-        if not isinstance(apply_zscore, ApplyZScore) and apply_zscore is not False:
-            raise ValueError(f"apply_zscore should be either a boolean or one of [{', '.join([e.value for e in ApplyZScore])}]")   # noqa: E501
+        try:
+            apply_zscore = ApplyZScore(apply_zscore)
+        except NotImplementedError as e:
+            raise NotImplementedError(
+                f"Unknown value for apply_zscore: {apply_zscore!r}; valid options: [{', '.join([e.value for e in ApplyZScore])}] or Bool (True defaults to {ApplyZScore.TRIAL})") from e  # noqa: E501
```
qubixes, line 190: Okay, I found out that in an f-string {a!r} means {repr(a)}. This seems a little bit obscure to me, the `repr` version will ring a bell with more people I think.

I'm also confused as to why not the ApplyZScore.parse() method is used.

Also, then it makes more sense to put the error message in the `parse` method and not catch and reraise?

qubixes: But perhaps I'm so confused I'm missing something.

hmp/transform.py
```diff
@@ -274,13 +309,14 @@ def __init__(
...
-        elif method is None:
+        elif method is AnalysisMethod.NO_ANALYSIS:
 
             data = data.rename({"channels": "component"})
             data["component"] = np.arange(len(data.component))
             data.attrs["pca_weights"] = np.identity(len(data.component))
 
-        if apply_zscore:
+
+        if apply_zscore.apply():
```
qubixes, line 319: You could use a `__bool__`  dunder to make it easier to read in this context I think. (I.e. you just need to rename the apply method to `__bool__`.)

qubixes (changes requested):
Looks better, just have some suggestions/questions on further improvement.

GWeindel:
I agree with the changes, and trust you both for the OOP parts. The tests are failing as the new implementation doesn't deal with `apply_zscore = False`

hmp/transform.py
maartenschermer: The absent `parse` was a bug.

GWeindel:
Looks all good to me, should we merge?

GWeindel:
One last thing, the `transform_data` should now be `preprocessing` as otherwise it is a bit confusing with the `transform` method from the `models` classes

hmp/transform.py
```diff
@@ -49,7 +49,7 @@ def __str__(self) -> str:
         return self.value
 
     def __bool__(self) -> bool:
-        return self != cls.NO_ANALYSIS
+        return self != cls.DONT_APPLY
```
GWeindel, line 52: This was a mistake, but somehow `method=None` or `method='no_analysis'` doesn't work

hmp/transform.py
```diff
@@ -68,7 +68,6 @@ def parse(cls, label):
             raise KeyError(f"Unknown method: '{label}'; valid options: {', '.join([e.value for e in cls])} or None")  # noqa: E501
 
 
-# TODO: move to utils
```
GWeindel: Should this still be done?

hmp/transform.py
```diff
@@ -102,7 +101,7 @@ def user_input_n_comp(data):
 # TODO: n_comp  None value input
 # TODO: printing to proper logging
 
-class DataTransformer:
+class Preprocessing:
```
GWeindel, line 104: Proposed new names, transform.py -> preprocessing.py (not yet implemented), DataTransformer -> Preprocessing

GWeindel (changes requested):

hmp/transform.py
GWeindel: I suggest we do that as a separate PR?

PR #186, opened 2025-04-24
n_events as mandatory parameter for FixedEventModel

GWeindel:
Up to now users could let it to None and the code would have inferred it during the fit but that doesn't make sense as a fixedn model is built on an assumed number of events


qubixes (approved):
Sounds like a good change, but I would also make the parameter itself mandatory and not optional.

GWeindel:
Made it mandatory and assert the type also to avoid weird error messages if non int types are passed

PR #187, opened 2025-04-25
floating point correction, plus capture fit failure

GWeindel:
I don't know if this situation (lkh = -np.inf) can occur after this fix, but I prefer to capture it as it is diagnostic of a failure

### DIFF
--- a/hmp/models/fixedn.py
+++ b/hmp/models/fixedn.py
@@ -258,23 +258,20 @@ class FixedEventModel(BaseModel):
             max_lkhs = np.argmax(lkhs)
         else:
             max_lkhs = 0
-            
-        if lkhs.sum() != -np.inf:
+
+        if np.isneginf(lkhs.sum()):
+            raise ValueError("Fit failed, inspect provided starting points")
+        else:
             self._fitted = True
             self.lkhs = lkhs[max_lkhs]
             self.magnitudes =  np.array(estimates[max_lkhs][1])
             self.parameters = np.array(estimates[max_lkhs][2])
             self.traces = np.array(estimates[max_lkhs][3])
             self.param_dev = np.array(estimates[max_lkhs][4])
-    
             self.level_dict = level_dict
             self.levels = levels
             self.mags_map = mags_map
             self.pars_map = pars_map
-        else:
-            raise ValueError("Fit failed, inspect provided starting points")
-        
-
 
     def transform(self, trial_data):
         _, levels, clabels = self.level_constructor(


hmp/models/fixedn.py
```diff
@@ -253,23 +253,28 @@ def fit(
...
-        self.parameters = np.array(estimates[max_lkhs][2])
-        self.traces = np.array(estimates[max_lkhs][3])
-        self.param_dev = np.array(estimates[max_lkhs][4])
-
-        self.level_dict = level_dict
-        self.levels = levels
-        self.mags_map = mags_map
-        self.pars_map = pars_map
+            
+        if lkhs.sum() != -np.inf:
```
qubixes, line 262: I would reverse the if statement to save some indentation, so `if <error_cond>: raise Error()`.

Also, for consistency you should probably use `np.isneginf`?

qubixes (changes requested):
Sounds like a reasonable fix, but I haven't seen the error of course. Just a small improvement suggestion.

hmp/models/fixedn.py
GWeindel: I implemented your suggestion and ready to merge!

GWeindel:
I added @jelmerborst as a reviewer as he also had some issues with these -inf. 
@Jelmer I figured that these come from trials with very short latencies in which the location was turning all probabilities to 0, changing the censoring of the location by just adding a non-zero minimal penalty (1e-15) seems to do the trick to avoid any -inf

JoKra1:
Hey, sorry I am late to the discussion!

Also sorry for being pedantic, but I still don't like this fix. I also don't think it addresses the underlying issue.

The probabilities collapsing to zero has at least two sources:

1) Mass becomes zero everywhere for very small scale parameters.
2) Mass becomex infinite at zero for some distributions (an exception is the Weibull) which have
dynamic support depending on the shape parameter. Thus, these distributions (Gamma, inverse Gaussian)
are theoretically only defined with support for duration > 0.

Both problems lead to ``model.distribution_pmf(shape, scale,max_duration)`` returning only zeros and thus to invalid probability distributions. In the second case this happens because during the normalization over ``model.pdf(np.arange(max_duration),shape,scale=scale)`` everything becomes ``nan`` due to the ``inf`` mass at index 0. I have created a [notebook](https://github.com/JoKra1/hsmm_mvpy/blob/fix_611/support_issue.ipynb) on a fork based on this exact commit for you to check this out visually.

In practice the first case is indeed likely the result of very short peak-2-peak latencies, resulting in very
small scale parameteter estimates. However, the hack to add 1e-15 is not sufficient to fix this for
two reasons:

- It does not prevent the possibility of total support collapse for stages without a
``location`` parameter.
- It is theoretically invalid for stages with non-zero ``location``: it says that all the mass is
allocated to impossible peak-to-peak distances ``0:locations[stage]``! This is not justified as discussed in #262. The ``gain`` computation implemented currently assumes no overlap between any two patterns and thus requires some peak-to-peak distances to be treated as impossible.

The real solution would be to ensure at each EM step that parameter estimates are constrained to
result in valid probability distributions. This can be achieved by halving the parameter update vector iteratively until the resulting step produces a valid (non neginf) llk. This is implemented in #267.

GWeindel:
Yes I agree your fix looks much better and thanks for taking time to clearly illustrate the problem. Would you mind doing a separate PR as this addresses a separate issue that is relevant also for scale-only distributions?

JoKra1:
Ya I am making a bit of a mess, sorry for that.. somehow these issues are still all related, maybe we should start a discussion thread where we can collect and coordinate things? Of course I can then still make a single new pr!

GWeindel:
I'm following no worries, if you think of a handy way to have these issues connected I'm all in but no need just for me

PR #189, opened 2025-05-02
Minor fixes on utils, visu, fixedn and cumulative

GWeindel:


qubixes (approved):
Don't see anything out of the ordinary, approved!

PR #193, opened 2025-05-12
Fix visu import

GWeindel:


qubixes (approved):
I'm fine with the change back too.

PR #194, opened 2025-05-14
Only use np.array in estimation

GWeindel:
xarray only used when outputting event probabilities.

Structure is clearer and slight speed advantage of the new version if going through a lot of EM iterations


GWeindel:
It was for optimization purposes but after extensive testing the gain is very weak if any... The downside is less consistency so I'm starting to question that PR. I will leave it hanging for now

PR #195, opened 2025-05-14
Create distribution class

GWeindel:
Doesn't work yet, just implemented the general idea of a class for distribution


GWeindel:
@qubixes can we make a call to fixedn model with default distribution and event classes? 

Something like:
```python
model = FixedEventModel(n_events=n_events, sfreq=epoch_data.sfreq)
```
Where `event_properties` and `distribution` are automatically called and set to defaults?

hmp/distributions.py
```diff
@@ -0,0 +1,43 @@
+class GammaDistribution():
+    def __init__(self, shape=2):
+        from scipy.stats import gamma  
+        self.shape = shape
+        self.pdf = gamma.pdf
+    def scale_to_mean(self, scale):
```
qubixes, line 6: Should have a new line between methods. Also the class and methods should have a docstring.

hmp/distributions.py
```diff
@@ -0,0 +1,43 @@
+class GammaDistribution():
+    def __init__(self, shape=2):
+        from scipy.stats import gamma
```
qubixes, line 3: Imports should be at the top of the module.

hmp/models/base.py
```diff
@@ -83,45 +34,14 @@ class BaseModel(ABC):
...
-
-                self.scale_to_mean, self.mean_to_scale = (
-                    _weibull_scale_to_mean,
-                    _weibull_mean_to_scale,
-                )
-            case _:
-                raise ValueError(f"Unknown Distribution {distribution}")
+        self.pattern = pattern
+        if distribution is None:
+            from hmp.distributions import GammaDistribution
```
qubixes, line 42: At the top of the module here as well.

hmp/patterns.py
qubixes, line 1: Looks like a good improvement!

hmp/simulations.py
```diff
@@ -346,21 +348,20 @@ def demo(cpus, n_events, seed=123):
...
+    from hmp.io import read_mne_data
 
     random_gen = np.random.default_rng(seed=seed)
 
     ## Parameters for the simulations
     frequency, amplitude = (
         10.0,
-        0.3e-7,
-    )  # Frequency of the transition event and its amplitude in Volt
+        1e-7,
```
qubixes, line 358: These magic numbers should at some point be defined somewhere, but not for this PR.

hmp/simulations.py
```diff
@@ -376,11 +377,11 @@ def demo(cpus, n_events, seed=123):
             [name_sources[source], frequency, amplitude, gamma(shape, scale=times[source])]
         )  # gamma returns times in ms
 
-    file = "dataset_tutorial2"  # Name of the file to save
+    file = "demo_dataset"  # Name of the file to save
 
     # Simulating and recover information on electrode location and true time of the simulated events
     files = simulate(
-        sources, n_trials, cpus, file, overwrite=False, seed=seed, noise=True, sfreq=sfreq
+        sources, n_trials, cpus, file, path='sample_data/', overwrite=False, seed=seed, noise=True, sfreq=sfreq
```
qubixes, line 384: Why the '/' after sample_data? Might not work with some Windows versions.

qubixes (changes requested):
Overall looks like a good improvement in the structure. Ideally, the events -> pattern change would be in its own PR, but of course for this one let's merge them in one go.

PR #196, opened 2025-05-21
Loocv

qubixes:
Some work on the loocv calculations. Unfinished.


qubixes:
@GWeindel We should probably just close this one, right?\

GWeindel:
Yes we can

PR #200, opened 2025-06-05
pass preprocessing object to TrialData

GWeindel:
also possible to pass Preprocessing().data


qubixes (approved):
Nothing really to comment on! Perhaps you could add the type hints while you're at it?

GWeindel:
Thanks! I'll add them with the docstrings otherwise I'll get lost

PR #204, opened 2025-06-16
Preprocess loose ends

maartenschermer:
+ Moved `user_input_n_comp` function to utils.
+ Removed redundant code from utils.


GWeindel:
Thanks Maarten. For the `zscore_xarray()` it should have started with an underscore so you can safely remove it too

GWeindel (changes requested):

PR #208, opened 2025-06-17
renamed a few functions and modules

GWeindel:


GWeindel:
The PR is relatively safe so I'll just merge

PR #209, opened 2025-06-17
Re-organized parameters between __init__ and fit()

GWeindel:
Also changed the use of base_fit to better work as users can expect

### DIFF
--- a/hmp/models/eliminative.py
+++ b/hmp/models/eliminative.py
@@ -25,6 +25,14 @@ class EliminativeMethod(BaseModel):
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


qubixes:
@GWeindel Can you explain why you have changed it? I feel that we have had this exact discussion before namely. Has something changed?

GWeindel:
@qubixes right we had that discussion on `EventModel`. Here I just removed `max_event` and `min_event` from `__init__` as they didn't have any effect but indeed it would be better to have all in `init` and nothing in `fit`. I'll correct.

PR #211, opened 2025-06-19
Rendering website

GWeindel:


README.md
```diff
@@ -209,14 +204,14 @@ epoch_data.sel(channel=['EEG 001','EEG 002','EEG 003'], sample=range(400))\
...
-        reject_threshold:  inf
+        upper_limit_rt:    inf
+        reject_threshold:  None
         n_trials:          200
 
 
 
     
-![png](README_files/README_12_1.png)
+![png](README_files/README_11_1.png)
```
qubixes, line 214: Perhaps it is an idea to give the figures a more descriptive name at some point?

README.md
```diff
@@ -315,13 +308,13 @@ plt.legend()
...
 
-    <matplotlib.legend.Legend at 0x751a2835a030>
+    <matplotlib.legend.Legend at 0x7d79e28347d0>
 
 
 
 
     
-![png](README_files/README_22_1.png)
+![png](README_files/README_21_1.png)
```
qubixes, line 317: Then also you don't need to rename them perhaps?

qubixes (approved):
It is hard for me to know exactly what to check in this PR. I have checked the README. Let me know if you want other parts to be checked as well.

README.md
GWeindel: Those are automatically generated so I wouldn't dare going into that

GWeindel:
If you can check the docs/source .py and .rst files that would be great as I mainly tried things by trial and error, but it does look good

GWeindel:
Should work now, also I assume "nbsphinx" had to be added in `[project.optional-dependencies]`

PR #213, opened 2025-06-20
Cleaned trial_data of redundant infos and adapted downstream

GWeindel:


qubixes (approved):
Looks very reasonable to me!

PR #228, opened 2025-07-17
Lint distributions.py

qubixes:
Oh, you have main as the default branch? Ah yes, I remember now. I am working with too many branches to remember. I will adjust.


GWeindel (approved):

GWeindel:
Do we lint through PR directly in main?

PR #229, opened 2025-07-17
Ruff io

qubixes:
@GWeindel I need your help with this one a bit. Also, I do think io.py might need some refactoring at some point. Some code doesn't seem to work (anymore) or otherwise has no effect, would be good if you could check that.


qubixes:
@GWeindel Can you take a brief look at this?

hmp/io.py
```diff
@@ -359,7 +370,7 @@ def read_raw_and_epoch(
             events = mne.find_events(
                 data, verbose=verbose, min_duration=1 / data.info["sfreq"]
             )
-        except:
+        except:  # Which exception?
```
GWeindel: It's a ValueError (https://github.com/mne-tools/mne-python/blob/maint/1.10/mne/event.py#L755-L762)

hmp/io.py
```diff
@@ -458,7 +469,7 @@ def _epoch_selection(epochs,
     if metadata is None:
         try:
             metadata_i = epochs.metadata  # accounts for dropped epochs
-        except:
+        except:  # Which exception?
```
GWeindel:
```suggestion
        except AttributeError:
```

hmp/io.py
```diff
@@ -482,7 +493,7 @@ def _epoch_selection(epochs,
             rts = metadata_i[rt_col]
         try:
             rts = rts / scale
-        except:
+        except:  # Which exception?
```
GWeindel: It actually doesn't make sense her, I'll suggest soemthing

GWeindel:
I removed the save_xr and load_xr as I think it's not useful given that pickling works fine

GWeindel (approved):

PR #231, opened 2025-07-17
Ruff simulations.py

qubixes:


hmp/simulations.py
```diff
@@ -19,11 +22,6 @@ def available_sources():
     return labels
 
 
-def sfreq():
-    """Recovering sampling frequency of the sample data."""
-    return info()["sfreq"]
```
GWeindel, line 24: This needs to be changed in the documentaiton as well, I'll deal with it in a bit

GWeindel: Nevermind it's good

GWeindel (approved):

PR #234, opened 2025-07-17
Lint utils.py

qubixes:


GWeindel (approved):
I agree with the refactoring, I'll also think of the too many branches in event_times

PR #235, opened 2025-07-17
Ruff visu.py

qubixes:


GWeindel (approved):
Also agree with the refactoring here...

PR #238, opened 2025-07-24
Fix models/eliminative.py

qubixes:


hmp/models/eliminative.py
```diff
@@ -10,12 +10,10 @@
 from hmp.models.event import EventModel
 from hmp.trialdata import TrialData
 
-
 try:
     __IPYTHON__
-    from tqdm.notebook import tqdm
 except NameError:
-    from tqdm import tqdm
+    pass
```
GWeindel: This will make a problem downstream when tqdm is called no?

GWeindel (changes requested):

hmp/models/eliminative.py
qubixes: But tqdm doesn't seem to be called at all?

GWeindel: Right, it was legacy from before the refactoring

GWeindel (approved):

PR #245, opened 2025-09-24
Implemented offset before stimulus

rickdott:
Change to allow adding additional samples before offset, in my usecase I later remove them from the dataset so they are not used in actual HMP estimation but I dont see why they couldn't be (perhaps when you want to examine prestim effects like anticipation)

Implementation is very simple, @GWeindel do you think this might mess with RT calculation somewhere? If you do RT * sfreq to find the samples it should now be shifted by offset_before, what do you think?

Also implemented a check for when selected time before stim is earlier than tmin of the epochs object, assuming this is **always** in seconds

### DIFF
--- a/hmp/io.py
+++ b/hmp/io.py
@@ -33,8 +33,8 @@ def read_mne_data(  # noqa: PLR0913  # This should probably be refactored instea
     verbose: bool = True,
     tmin: float = -0.2,
     tmax: float = 5,
-    offset_after_resp: float = 0,
-    offset_before_stim: float = 0,
+    offset_start: float = 0,
+    offset_end: float = 0,
     high_pass: float | None = None,
     low_pass: float | None = None,
     pick_channels: str | list = "eeg",
@@ -109,10 +109,10 @@ def read_mne_data(  # noqa: PLR0913  # This should probably be refactored instea
         Start time (in seconds) relative to stimulus onset for epoching.
     tmax : float, default=5
         End time (in seconds) relative to stimulus onset for epoching.
-    offset_after_resp : float, default=0
-        Additional time (in seconds) to include after the response onset.
-    offset_before_stim : float, default=0
-        Additional time (in seconds) to include before the stimulus onset.
+    offset_start : float, default=0
+        Additional time (in seconds) to include before the epoch starts.
+    offset_end : float, default=0
+        Additional time (in seconds) to include after the end of the epoch.
     high_pass : float, optional
         High-pass filter cutoff frequency.
     low_pass : float, optional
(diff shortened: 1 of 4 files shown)


hmp/preprocessing.py
```diff
@@ -358,6 +360,7 @@ def __init__(  # noqa  # Might need to be refactored.
 
         data.attrs["sfreq"] = sfreq
         data.attrs["offset"] = offset
```
GWeindel: Change t
```suggestion
        data.attrs["offset_end"] = offset_end
```

hmp/preprocessing.py
```diff
@@ -358,6 +360,7 @@ def __init__(  # noqa  # Might need to be refactored.
...
+        data.attrs["offset_before"] = offset_before
```
GWeindel: Clearer I would say, also for people who might use other things than stimulus and response as interval
```suggestion
        data.attrs["offset_start"] = offset_start
```

hmp/io.py
```diff
@@ -642,7 +656,9 @@ def hmp_data_format(
     events : np.ndarray, optional
         Description for each epoch and participant that need to be stored (e.g. condition)
     offset : float, default=0
-        Offset in seconds to apply to the data.
+        Offset in samples to apply after the response time.
+    offset_before : float, default=0
+        Offset in samples to apply before the stimulus onset.
```
GWeindel: Is this negative then if we want samples before stimulus?

GWeindel (changes requested):
Nice! About the RT I'm not completely sure, tests are passing so that's already good but not sure all the situations are covered... I can take a look but it will require some time

hmp/preprocessing.py
rickdott: I kept the pre-existing offset as 'offset' to not break any other code that uses this term (including my own). Do you prefer changing this? We could also specify that 'offset' is at the end in the docstring. Let me know which you prefer and I'll adjust accordingly.

rickdott:
Adjusted naming, also looked for '.offset' in the HMP repo and adjusted where necessary. Name is now 'offset_end'. Tested by fitting a small HMP and visualizing it.

@GWeindel do you want me to adjust the user-facing (read_mne_data) parameter 'offset_start' to use negative numbers? I think using negative numbers now would shorten the epoch.

GWeindel:
This looks good, I fear though that the structure changes in #247 break your changes. Would you consider applying this to #247 after it has been merged (matter of days)? In this new version you don't need to change the input functions and instead only change the transformer operations (yes I insist on that name :))

rickdott:
Of course!

Issue #57, opened 2023-06-22
Testing needed for the new ways of zscoring


GWeindel:
PR #54 introduced a new default way of scoring the data. The previous behavior of the hmp.transform_data() function was to zscore on a by trial basis, the new default way zscore the data by-participant, a third way is to score all the data at once.

It is yet unclear what the best default option is. Further testing is needed to validate the default behavior

GWeindel:
I tested this issue by re-running Figure 3 of the preprint (see caption of the paper).

This is the original result (zscore per participant/all in the case of only one participant per dataset):
![amp-tol-nev](https://github.com/GWeindel/hmp/assets/28750455/9f475591-2b53-4456-957a-d5a5e306cf93)

This is when szcoring each trial:
![amp-tol-nev_pertrial](https://github.com/GWeindel/hmp/assets/28750455/c9cb5b2e-e505-4edf-b6ac-30e657cc6ef6)

This is without zscoring:
![amp-tol-nev_general-szcoring](https://github.com/GWeindel/hmp/assets/28750455/4f007d05-2ec4-49b5-a15b-2a1a66265ee8)

Overall it is clear that zscoring per trial has the best performance at least in the case of one dataset (one participant one condition). It maybe that zscoring this way can change condition differences, this is still to be tested but I suggest that we move to this default as it also improve things on real data

jelmerborst:
I agree. I would also make zscore_acrossPCs the default: https://github.com/GWeindel/hmp/blob/increasing_location/src/hsmm_mvpy/utils.py#L550.

GWeindel:
Right, I'll try that also

GWeindel:
Can I just take that part from your branch so I don't press you too much to be ready for a PR? :)

GWeindel:
It looks like doing this zscoring across PCs is making things worse
![amp-tol-nev_pertrial_acrossPCs](https://github.com/GWeindel/hmp/assets/28750455/4c7ca2a5-5d52-40bb-b0ea-7e2436593fa0)
I can also output the models coefficients but I think graphically it's enough to say that in the case of these simulations it's not a good strategy. I tried on real data and this also seems to worsen things.

My take on this is that it's not a problem if PCs have the same variance as the magnitudes of HMP can pick or not on each PCs (see matrix plot of tutorial 1). However having very low variance components can be problematic in the estimation procedure.

PR #247, opened 2025-11-10
Refactoring io preprocessing

GWeindel:
Hi @jelmerborst , @lvmaanen @rickdott @qubixes @maartenschermer 
I made a quite big PR (oups). I guess none of you has the time to read it all so I just tag each of you on the parts that will probably interest you/needs your review.

# Major changes
1. The EEG/MEG data is now stored for the whole epochs instead of being cut at RT when the data is read using `read_mne_data`. That allows to more easily plot ERPs but also switch between different durations without re-reading all the data  @jelmerborst, @rickdott
2. The `preprocessing` has been renamed to `transformers` to make the distinction with EEG/MEG preprocessing, typically done before HMP. The new classes follow the modular organization as in `models` @qubixes @maartenschermer
3. The filtering option was removed from the transformers as there are a few users who were using it wrongly at this stage and the maintenance is tricky @jelmerborst 
4. Cutting the data at the RT (or more generally `interval` now) is performed as part of the new `transformers` thus the arguments that applied to this operation were moved and went from `read_mne_data` to the `transformers` classes. 
5. Zscoring is now always performed at the trial level without any other option (amongst others to avoid large RAM usage) @lvmaanen @jelmerborst This is because we expect the same sequence in all trials, thus that's the only reasonable transformation given this assumption. Note that this does not mean that we remove the difference, e.g. the channel values at the events' time (typically computed using the original epoched data) won't be affected by this zscoring strategy. Furthermore Issue #57 shows that this is the best strategy for simulated data at least (and real data from my experience)
6. The PCA is now computed without sklearn to avoid the dependency and is applied on the average covariance matrix with a small regularization. These changes respectively gives more control and slightly improves performance
7. MCCA was removed for now as it is too experimental but might be re-implemented later on @jelmerborst 
8. The result of preprocessed data is a stacked representation with trials X components X samples instead of all_samples X components which wasn't easy to use for others.

# Overall consequences:
- The interval cut (as well as providing metadata) is now optional and the data read is the actual data stored as epochs
- The distinction between preprocessed data and trial data is much clearer. 
- The different projections method now share all the same structure
- The transforming step now uses 2x less RAM and is also slightly faster (mostly due to the change of `float64` to `float32` and the removal of large array operations in `transformers`)

# Coming 
In another PR I will further refactor the `io` module to allow for more flexibility in the data type being read and also to avoid very long functions with unclear paths


GWeindel:
The tests are broken, I'll fix when we agree on the new revision and add some tests to the new structure

hmp/transformers/common.py
```diff
@@ -0,0 +1,79 @@
+"""Transforms epoched data using a a custom linear combination for HMP analysis.
```
qubixes:
```suggestion
"""Transforms epoched data using a custom linear combination for HMP analysis.
```

hmp/transformers/custom.py
qubixes, line 1: Change filename to lowercase (drop the proj?).

hmp/transformers/common.py
```diff
@@ -0,0 +1,79 @@
...
+    ----------
+    epoch_data : xr.Dataset
+        Input EEG data with dimensions [participant, epoch, sample, channel], from `io` module
+    weights: xr.DataArray
+        Custom linear combination of channels as an xarray.DataArray with 'channel' and 'component' dimensions and the weights
+    interval_id: str
+        Name of the variable that contains the per-trial intervals in the epoch_data used for cropping.
+    offset_after : float
+        Time offset after interval start for cropping.
+    too_short : float, optional
```
qubixes: You might want to consider the naming of this variable. I would think it is a boolean. Perhaps min_duration or something like it?

hmp/transformers/common.py
```diff
@@ -0,0 +1,79 @@
...
+        Minimum duration threshold for keeping epochs.
+    too_long : float, optional
```
qubixes: Same here.

hmp/transformers/ProjCustom.py
```diff
@@ -0,0 +1,79 @@
...
+        Time offset after interval start for cropping.
+    too_short : float, optional
+        Minimum duration threshold for keeping epochs.
+    too_long : float, optional
+        Maximum duration threshold for keeping epochs.
+    reject_threshold : float, optional
+        Threshold for rejecting noisy epochs.
+    apply_standard : bool
+        Whether to standardize variance across participants.
+    apply_zscore :bool
```
qubixes:
```suggestion
    apply_zscore : bool
```

hmp/transformers/identity.py
qubixes, line 1: Naming

hmp/transformers/ProjIdentity.py
```diff
@@ -0,0 +1,86 @@
...
+        Time offset after interval start for cropping.
+    too_short : float, optional
+        Minimum duration threshold for keeping epochs.
+    too_long : float, optional
+        Maximum duration threshold for keeping epochs.
+    reject_threshold : float, optional
+        Threshold for rejecting noisy epochs.
+    apply_standard : bool
+        Whether to standardize variance across participants.
+    apply_zscore :bool
```
qubixes:
```suggestion
    apply_zscore : bool
```

hmp/transformers/identity.py
```diff
@@ -0,0 +1,86 @@
...
+            too_long=too_long,
+            reject_threshold=reject_threshold,
+            verbose=verbose,
+            apply_standard=apply_standard,
+            apply_zscore=apply_zscore,
+            centering=centering,
+            copy=copy,
+        )
+        warn('Identity projection might pose problems of dimensionality'
+             'and collinearity of channels. Thus rendering HMP estimation'
```
qubixes:
```suggestion
             ' and collinearity of channels. Thus rendering HMP estimation'
```

hmp/transformers/identity.py
```diff
@@ -0,0 +1,86 @@
...
+             'difficult, use with care!')
```
qubixes:
```suggestion
             ' difficult, use with care!')
```

hmp/transformers/pca.py
```diff
@@ -0,0 +1,137 @@
...
+        eigvals, eigvecs = eigh(pca_ready_data)
+        ix = np.argsort(np.abs(eigvals))[::-1]
+        evecs = eigvecs[:, ix]
+        evecs = evecs[:, :n_comp]
+        # Rebuilding as xarray to ease computation
+        coords = dict(channel=("channel", channel), component=("component", np.arange(n_comp)))
+        pca_weights = xr.DataArray(evecs, dims=("channel", "component"), coords=coords)
+        return pca_weights, eigvals[ix]
+
+    def user_input_n_comp(self, data, n_comp, channels):
```
qubixes: Data types?

hmp/transformers/base.py
```diff
@@ -0,0 +1,198 @@
...
+    
+    
+        Returns
+        -------
+        epoch_data : np.ndarray
+            Array of cropped epoch data that passed all criteria.
+        """
+        sfreq=self.sfreq
+        too_short = self.too_short
+        too_long = self.too_long
```
qubixes: Is this really necessary? Might be clearer if you just use self.too_short.

hmp/transformers/base.py
```diff
@@ -0,0 +1,198 @@
...
+            warn(f"Found intervals with a max value value of {np.round(max_rt,2)}"
+                ", assuming intervals are in milliseconds and converting to seconds")
+            rts_arr /= 1000
+        offset_after_samples = int(np.rint(self.offset_after * sfreq))
+    
+        if too_long is None:
+            too_long = (int(epoch_data.sample.max()) - offset_after_samples)/sfreq
+        if too_short is None:
+            too_short = 1 / sfreq
+        if too_long < 0 or too_short < 0 or too_long < too_short:
```
qubixes: This should be validated during `__init__`.

hmp/transformers/base.py
```diff
@@ -0,0 +1,199 @@
...
+                elif ~np.isnan(epoch_data.values[i, 0, time0]):
+                    epoch_data.values[i, :, :] = np.nan
+                    rej += 1
+                else: # assumes rejected before
+                    epoch_data.values[i, :, :] = np.nan
+                    inexistant_rej += 1
+            else:
+                epoch_data.values[i, :, :] = np.nan
+    
+        assert rej < len(epoch_data), 'All trials rejected, inspect intervals and rejection criterion'
```
qubixes: Don't use asserts for non-testing code. Also, shouldn't it be the other way around?

qubixes (changes requested):
I have a few small comments, looks quite nice overall!

hmp/transformers/base.py
GWeindel: It was correct but probably the assert syntax made this unclear

GWeindel:
I updated the code based on @qubixes feedback and made passing tests so it's production ready. Note that I also removed zscoring completely and only "whiten" the components by dividing them with their std computed on all samples. @rickdott @jelmerborst @lvmaanen  any additional comments or suggestions? I'm thinking to merge on monday

rickdott:
I'll look into the part you tagged me in tomorrow, quick check: did you also update the tutorials/docs to use the new preprocessing/transformers API? Dont see it in the changed files

GWeindel:
> I'll look into the part you tagged me in tomorrow, quick check: did you also update the tutorials/docs to use the new preprocessing/transformers API? Dont see it in the changed files

I updated the tutorials, to look at the API you need to build it
```bash
pip install -e ".[docs]"
cd docs; make clean html
```
And then open the index.html in docs/build/html.

hmp/transformers/pca.py
```diff
@@ -0,0 +1,156 @@
...
+        If None (default), a prompt will open asking to select a number
+    """
+
+    def __init__(#noqa: PLR0913
+        self,
+        epoch_data: xr.Dataset,
+        interval_id: str = 'rt',
+        offset_after_end: float = 0,
+        min_duration: float = 0,
+        max_duration: float =float('Inf'),
```
rickdott: Forgot a space here

rickdott: max_duration: float = float('Inf'),

hmp/transformers/base.py
```diff
@@ -0,0 +1,208 @@
...
+
+from abc import ABC
+from typing import Optional
+from warnings import warn
+
+import numpy as np
+import xarray as xr
+
+
+class BaseTransformer(ABC):
```
rickdott, line 37: I dont 100% see the use of this being a class rather than separate functions, maybe it could be changed to take epoch_data in its initialization, we could then run self.common_preprocess() on that within the BaseTransformer.__init__(), since we run that function in every child of BaseTransformer

hmp/transformers/base.py
```diff
@@ -0,0 +1,208 @@
...
+    """Base class for HMP transformer pipelines.
+
+    This class provides common transformer steps.
+
+    Methods
+    -------
+    common_preprocess()
```
rickdott: common_preprocess also takes a data argument (currently, see prev comment)

hmp/trialdata.py
```diff
@@ -43,35 +44,31 @@ class TrialData:
...
+    def from_transformer(cls, transformed, pattern, dtype = np.float32):
         """
-        Create a TrialData instance from preprocessed data and a given pattern.
+        Create a TrialData instance from transformed data and a given pattern.
 
         Parameters
         ----------
-        preprocessed : Preprocessing or xr.DataArray
-            The preprocessed object or xarray DataArray containing the preprocessed data.
+        transformed : BaseTransfromer or xr.DataArray
```
rickdott, line 55: typo (BaseTransfromer)

rickdott (commented):
Checked io & transformers and left some comments (not as a review, sorry, hope you can find them). In general it is probably a good change, but I didn't mind the use of 'preprocessing' since that module did do some things I relate to EEG preprocessing (bandpass filters, removing bad trials). Transformers to me sound too much like the robots or Vaswani (2017), this is just semantics though.

GWeindel:
@rickdott I finished, documentation is also updated, I only need to rework on the tests but you can try if you have time, default should be fine

GWeindel:
The PR is ready to be integrated

Of note is that the default behavior changed slightly compared to what I wrote in the beginning, by default data is: 	
	- centered (mainly for covariance computation) 
	- the trials are standardized (across components) -> All trials are contributing equally
	- the participants are zscored (over components) -> All components are re-scaled for each participant as they might not all have the same SNR
	- SVD is explicitly used rather than PCA (but was already the case before this PR as it was the option passed to sklearn)

rickdott:
Looks good, I dont think there are review-worthy changes in the commits since I last took a look.

rickdott (approved):

PR #250, opened 2026-01-21
Added offset before start of interval and adjusted naming/docs

rickdott:
As requested, added offsets before interval start to the new IO update.
Props for making it easier :-)

### DIFF
--- a/hmp/transformers/base.py
+++ b/hmp/transformers/base.py
@@ -61,6 +61,7 @@ class BaseTransformer(ABC):
             self,
             interval_id: str,
             offset_after_end: float,
+            offset_before_start: float,
             min_duration: Optional[float],
             max_duration: Optional[float],
             reject_threshold: Optional[float],
@@ -73,6 +74,7 @@ class BaseTransformer(ABC):
     ):
         self.interval_id = interval_id
         self.offset_after_end = offset_after_end
+        self.offset_before_start = offset_before_start
         self.min_duration = min_duration
         self.max_duration = max_duration
         self.reject_threshold = reject_threshold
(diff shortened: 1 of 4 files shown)


rickdott:
Currently, providing a negative number removes those samples from the start, This is the only thing I'm uncertain about. Options:

- Keep as is (the docstring says the epoch is extended by the given amount before start)
- Take the absolute and don't support removing the start of intervals
- Reverse the behavior (-offset = offset before start, +offset = offset after start)
@GWeindel what do you think?

hmp/transformers/pca.py
```diff
@@ -30,7 +30,9 @@ class ProjPCA(BaseTransformer):
     interval_id: str
         Name of the variable that contains the trial intervals in the epoch_data used for cropping.
     offset_after_end : float
-        Time offset after interval start for cropping.
+        Time offset after interval end for cropping.
+    offset_before_start : float
+        Time offset before interval start for cropping. Epoch is extended by this amount before start.
```
GWeindel: Don't forget to also adapt in the other classes... Yes there must be a better way to link the docstrings

hmp/transformers/base.py
```diff
@@ -158,11 +161,14 @@ def reject_crop_epochs(self, epoch_data:xr.Dataset):
         rej = 0
         reject_threshold = self.reject_threshold if self.reject_threshold is not None else np.inf
         time0 = np.argmin(np.abs(epoch_data.sample.values))
+        if time0 - offset_before_start_samples < 0:
```
GWeindel: This will probably send a linting error for too many branching but leave it and I'll see what can be done

GWeindel (approved):

GWeindel:
This looks good. On the behavior I think using absolute makes the most sense, so -100 takes 100 ms before stim while +100 starts 100 after stim. That way it's coherent with MNE and the read_mne_data function

GWeindel:
Maybe while we're at it we should just change `offset_after_end` to `offset_end` and same for start as `offset_after_x` suggest that there is also `offset_before_x`. 
P.-S to clarify I meant absolute in time not the absolute of the value as you said in your first message

GWeindel (approved):

PR #254, opened 2026-03-09
For BIDS dataset allows to apply a preprocessing script

GWeindel:
User provides a path, e.g.:

```python
epoch_data = hmp.io.read_mne_data([],
                                  data_format='bids',
                                  sfreq=sfreq,
                                  bids_parameters={
                                      'bids_root': bids_root,
                                      'task': 'LRP',
                                      'session': None,
                                      'datatype': 'eeg',
                                  },
                                  preprocessing_path=os.path.join(bids_root,'code/preprocessing.py')
                                  )
```

with a set of instructions for preprocessing before epoching, e.g. preprocessing.py:

```python
def preprocessing_code(raw, events):
    raw = raw.pick_types(eeg=True)
    raw = mne.preprocessing.compute_current_source_density(raw)
    return raw

```

@qubixes In principle is this OK from a python perspective? (@rickdott might also be interested)

### DIFF
--- a/hmp/io.py
+++ b/hmp/io.py
@@ -13,6 +13,7 @@ import json
 import os
 import warnings
 from pathlib import Path
+from typing import Callable, Optional
 
 import mne
 import numpy as np
@@ -38,8 +39,8 @@ def read_mne_data( # noqa: PLR0913,PLR0912
     pick_channels: str | list = "eeg",
     reference: str | None = None,
     bids_parameters: dict = {},
+    preprocessing_fn: Optional[Callable] = None,
     dtype: DTypeLike = np.float32,
-    preprocessing_path: str | None = None,
 ) -> xr.Dataset:
     """Read EEG/MEG data format (.fif or .bdf) using MNE's integrated function.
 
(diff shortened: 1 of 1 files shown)


GWeindel:
I see a linting error because I guess the way I do it is weird:
```
F821 Undefined name `preprocessing_code`
   --> io.py:409:16
    |
407 |     if preprocessing_path:
408 |         run_preprocessing_script(preprocessing_path)
409 |         data = preprocessing_code(data, events)
    |                ^^^^^^^^^^^^^^^^^^
410 |
411 |     if metadata is None:

```
Have to think about that

rickdott:
Since read_mne_dara is already user facing I think it's a lot easier to just have the user provide a function (preprocessing_fn) and a preprocessing_kws (kwargs) dict. You could then do `preprocessing_fn(raw, events, **preprocessing_kws)`. I do something similar here: https://github.com/rickdott/hmp-ai/blob/main/src/hmpai/stagefinder.py in init and fit_model. In previous HMP versions I even had the fitting function as a parameter.

GWeindel:
That indeed does make a lot of sense, will implement thanks!

GWeindel:
@rickdott I added the function as an argument, I don't see the point yet to have kwargs in this case but I might come back to that later

PR #259, opened 2026-04-20
Large clean-up of cumulative method

GWeindel:
For maintenance and readability,  + changed name (and behavior) of `by_sample`

- By default the `step` is = to the location which makes sense given that this is the minimum duration (for now)
- Simplified termination rule
- Replaced `by_sample=False` with `fastforward = True` as name is more explicit
- The shortcut of looking for the max explored time (when  `fastforward = True`) as next starting point is used both if proposal was rejected or accepted
- When using  `fastforward = False` (so `by_sample=True `) do not store the channel parameters as their is an ambiguities as to which event these should be assigned to and we may trap the search algorithm because of an ill-defined starting point. Thus just store time pars and start fresh on channels
- Only store the submodels for accepted proposition (before accepted also the last solution, which didn't make sense)
- Removed `final_model_tolerance` as users can always refit with higher tolerance if that was ever useful (I added it but never actually used it)
- made the step (`j`) coherent when using `fastforward` True or False (always indexes progress)
- TQDM progress bar indexes the steps instead of the samples
- Output while fitting gives the array of times where events were found instead of the index of the last event
- Removed max_event_n as not useful and confusing
- improved docstrings following all these changes

### DIFF
--- a/hmp/models/cumulative.py
+++ b/hmp/models/cumulative.py
@@ -219,6 +219,7 @@ class CumulativeMethod(BaseModel):
             # look between which event the next proposition should go
             n_event_j = np.argwhere(scale_j >= np.cumsum(time_pars_props[:, 1])) + 2
             n_event_j = np.max(n_event_j) if len(n_event_j) > 0 else 1
+            n_event_j = np.min([n_event_j, n_events])  # do not insert even after last stage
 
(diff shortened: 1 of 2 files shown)


GWeindel:
@jelmerborst if you have time I'd love your feedback on that

Issue #262, opened 2026-04-23
Ensure pattern.location >= pattern.width for EventModel


JoKra1:
Hey Gabriel 👋🏻,

**Problem**: Patterns must not overlap for the likelihood computation to be correct. The ``patterns`` classes all expose a ``location`` parameter that can be used to ensure this, but no warning/error is raised when ``location`` is chosen smaller than ``width``, the number of samples taken up by the pattern.

**Proposed Solution**: Let the ``EventModel`` constructor raise an informative error ``if self.pattern.location < self.pattern.width and n_events > 1``. Overlap only becomes a problem with ``n_event > 1`` and implementing this check in the ``EventModel`` rather than the baseclass, allows for different model/estimation implementations in the future that do not have this constraint.

Below follows a more formal/extensive explanation of the problem, I tried wrapping it into a ``<details>`` block to hide it, but that messed up the math.. sorry for that!

**Formal Explanation**: The ``gain`` variable of a specific event $j$, computed for example in line 751 of the ``event.py`` model file, is the log of the likelihood ratio, i.e., the density of the signal given the onset of an event and current parameters divided by the density of the signal given no event onset.

More formally, let $\mathbf{c}_t$ be the cross correlation between the $K$ signals at time-point $t$ (i.e., electrodes, pcs, etc.) and the ``pattern``. Then, assuming the various normalizations computed as in the current implementation **and that patterns cannot overlap**, the density of observing $\mathbf{c}_t$, given that an event happens at $t$ and current patterns parameters $\mathbf{b}_j$ (also a $K$ vector, holding the weights for the pattern for every channel $k$ for event $j$), is equal to

$$
p(\mathbf{c}_t|\text{event}\ j\ \text{at }t, \mathbf{b}) = exp\left(T-\left\lbrace\sum_{k=1}^K(\mathbf{c}_{t\,k} - \mathbf{b}_{j\,k})^2\right\rbrace/V\right),
$$

where $T$ and $V$ are normalizing and scaling constants. Basically, since the $K$ signals are assumed to be mutually independent, the aforementioned density is proportional to the negative sum of the squared differences between the cross-correlated signal $\mathbf{c}_{t\,k}$ at channel $k$ and it's expected value ($\mathbf{b}\_{j\,k}$). Similarly, the density of observing $\mathbf{c}_t$, given that **no** event happens at $t$, is equal to

$$
[...]

PR #263, opened 2026-04-23
Addresses #262

JoKra1:
See #262 for discussion.

### DIFF
--- a/hmp/models/event.py
+++ b/hmp/models/event.py
@@ -81,9 +81,8 @@ class EventModel(BaseModel):
         self.channel_map = np.zeros((1, self.n_events))
         super().__init__(*args, **kwargs)
 
-        assert self.pattern.location >= self.pattern.width or n_events == 1, \
-         (
-             "For n_event > 1, pattern.location must be greater or equal than pattern.width"
+        if n_events > 1 and self.pattern.location < self.pattern.width:
+             warn("For n_event > 1, pattern.location must be greater or equal than pattern.width"
              f" but received pattern.location ({self.pattern.location}) is smaller than"
              f" but received pattern.width ({self.pattern.width})."
          )


hmp/models/event.py
```diff
@@ -81,6 +81,13 @@ def __init__(
         self.channel_map = np.zeros((1, self.n_events))
         super().__init__(*args, **kwargs)
 
+        assert self.pattern.location >= self.pattern.width or n_events == 1, \
+         (
+             "For n_event > 1, pattern.location must be greater or equal than pattern.width"
+             f" but received pattern.location ({self.pattern.location}) is smaller than"
+             f" but received pattern.width ({self.pattern.width})."
+         )
```
GWeindel: I would rather turn this into a warning for now at least
```suggestion
        if n_events > 1 and self.pattern.location < self.pattern.width:
             warn("For n_event > 1, pattern.location must be greater or equal than pattern.width"
             f" but received pattern.location ({self.pattern.location}) is smaller than"
             f" but received pattern.width ({self.pattern.width})."
         )
```

JoKra1: Oh yes, a warning might indeed be better, sorry I just went with assert because I saw you use that a couple lines earlier. I would change the language then from "must be greater or equal than" to "should be greater or equal than [...] otherwise parameter estimates might be biased." or something like that.. "must" really makes it sound like it needs an ``assert`` in my opinion 😬

> If you agee I would then move this to patterns,py in order to avoid this warning being triggered at every model init

Just to check, any model ``init`` will still initialise a new ``pattern`` by default (right?), unless the user specifies one, but the warning will not be raised by the default parameters? Then this sounds like a great idea, because I imagine that anyone using a custom ``pattern`` (which could potentially trigger the warning) would then just initialise that once at the top of the script and pass it to all subsequent models.

Should we preserve our discussion somewhere more permanent than #262 and refer users to that place as part of the warning? Because with the warning in place #262 can simply be closed.

GWeindel: The pattern is crosscorrelated with the data prior to the model creation, at the creation of trial-data, so `EventModel` does not initialize a 1pattern1 .

I agree with preserving the disscussion, I made a reference in the PR to your fork

JoKra1: Perfect, do you need me to click anything or can you close/merge/finalise as you see fit? Sorry, this is actually the first time where I not just immediately merge a pull request 😬.

GWeindel: I think you need to first accept my PR on your fork, which will update this PR and then you can merge

GWeindel:
Hi Josh, is there a particular reason why you did not integrate the PR that I made on your fork? That's the one containing the move to the `patterns.py` and the reference to the issue. But I can make a separate PR no worries

JoKra1:
Hey sorry if I messed something up, definitely wasn’t on purpose! But it’s weird, I did not see that you made a pull request to my fork (I just checked and it tells me there aren’t any open), all I saw was the changes you requested on my PR which I accepted 🤔.

I was a actually a bit confused because you mentioned that you wanted to move it to patterns, which made total sense, but did not see that in the changes 😬. But I just figured you might have decided differently, again sorry!!

So: no good reason, please move the warning to pattern init - it’s much better that way!

PR #265, opened 2026-04-23
Run release workflow only from core repo + node 20 deprecation bumps

JoKra1:
Small fixes @GWeindel, ensuring that:

- the release workflow will really only be triggered from your repo (it will generally fail from other repos since pypi will reject, so it's wasted computation)
- all actions using node.js v 20, which is deprecated and will soon be dropped, have been upgraded to use newer node versions.

Something to think about: I like that tests and lint checks automatically run in my fork, but not everyone might agree. So feel free to add ``if: github.repository == 'GWeindel/hmp'`` to other workflows as well if you think it's necessary.

### DIFF
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -15,9 +15,9 @@ jobs:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
     runs-on: "ubuntu-latest"
     steps:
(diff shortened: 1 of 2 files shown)


GWeindel:
Thanks! Right we could restrict the tests and lints but I too prefer to have them on my forks so I would leave it in

PR #267, opened 2026-04-24
Reference implementation EM step for scale and shape parameter of state distributions

JoKra1:
Hey @GWeindel & @jelmerborst ,

Sorry for the spam, but I have one final contribution before the weekend. 😬

**Purpose:** I wanted to make a **reference implementation for joint scale and shape estimation** of the duration distributions. For this, I chose to implement the standard EM step, since this is the algorithm currently used for everything else. This is not the only way to estimate both parameters, but it offers a good baseline for other approaches (more on that in a minute). If implemented correctly it provides MLE estimates, which is assuring and again makes it a good baseline for future improvements/other algorithms.

**Current issues:**
- It's much slower. This is mainly because of how the E-step output needs to be computed, and I just could not find a way quickly to handle the consoring with vectorized xarray numpy functions. So I used a nested loop structure. See the new ``estim_d_probs`` method. My strategy in cases like this would be to move the loop to C/C++. But I was not sure whether you're willing to commit to a dedicated build process Gabriel. If you are, I would be happy to show you how to set this up with github actions.
- ~There was a bug in ``distribution_pdf``, which computed the pmf for a duration of zero. 1) States cannot have zero duration in the current implementation and 2) some distributions, like the Gamma, assume support on the closed interval (0,infty], so zero excluded. Funny: the Gamma actually has variable support depending on the shape parameter, for a shape fixed at 2 this was never an issue, but now that the shape varies it caused problems. Fixed this, but might cause problems elsewhere..~
- Currently I just use scipy to optimise via quasi Newton. This would be a lot faster if we would provide the analytic derivatives for the duration distributions. If you're ok with imposing this extra theoretical hurdle on users wanting to implement a new duration distribution let me know, then I will see whether I can find some time to implement them.

**Math & Discussion of alternative implementations:**

First the math. It's a standard EM step, really. For the EM step we first compute the conditional probability distribution (essentially a Bayesian posterior) of the latent variables given the data and parameter estimates. For the HMP implementation there are two ways to think about this distribution. Since the model is strictly left to right, first (flat) state starts with time-point zero, and last (flat) state ends with last time-point, we can consider the distribution of event $j$ having onset $o_{s\,j}$ at any time-point $t <= T$, i.e., $p(o_{s\,j} = t | \mathbf{C}, \boldsymbol{\theta})$ on trial $s$, where $T$ is the maximum time-point across trials/subjects or we can (equivalently) consider the distribution of stage $i$ having duration $\tau_i=d$, i.e., $p(\tau_{s\,j} = d | \mathbf{C}, \boldsymbol{\theta})$ on trial $s$.

Unfortunately, to make the EM step work for joint scale & shape estimation, we need the second variant, but HMP currently only computes and returns the first one (in ``eventprobs``). Fortunately, given the model assumptions the second one can be computed from the first. Let $N$ denote the number of stages. We have two edge cases: the first and last stage. They have clear bounds (as described above) and so, for example, the probability of the first stage lasting for duration $d$ on trial $s$ is simply

$$p(\tau_{s\,1} = d | \mathbf{C}, \boldsymbol{\theta}) = p(o_{s\,1} = d | \mathbf{C}, \boldsymbol{\theta}),$$
 
Similarly, for the last stage $N$

$$p(\tau_{s\,N} = d | \mathbf{C}, \boldsymbol{\theta}) = p(o_{s\,N-1} = T_s - d | \mathbf{C}, \boldsymbol{\theta})$$

For stages in-between two events we need the joint probability of the onset of these two events:

$$p(\tau_{s\,j} = d | \mathbf{C}, \boldsymbol{\theta}) = \sum_{t=1}^{T_s}p(s\,o_{j-1} = t | \mathbf{C}, \boldsymbol{\theta}) * p(o_{s\,j} = t + d | \mathbf{C}, \boldsymbol{\theta}).
$$

The sum is necessary to marginalize over the specific onset moment, since different onset moments can still result in the same duration $d$. Frustratingly, that inner product needs to be set to zero if $t + d >= T_s$ to account for the aforementioned censoring - and since $T_s$ differs between trials I did not find a way to implement this efficiently.

Given these probabilities we are almost ready to turn to the M step. For the M step, one defines the $Q(\boldsymbol{\theta}|\boldsymbol{\theta}^k)$ function; the expectation of the "complete data" log-likelihood taken with respect to the posterior distribution of the latent variables given data $\mathbf{C}$ and parameters $\boldsymbol{\theta}$. Importantly, the posterior is treated fixed for this step **not as a function of the parameters**! Before we can define $Q(\boldsymbol{\theta}|\boldsymbol{\theta}^k)$ for HMP models, it is necessary to consider the assumptions of the model to make the function tractable:

- All $S$ **series are treated as mutually independent** given the model (parameters, i.e., $\boldsymbol{\theta}$).
- For the HMP model we assume the durations $\tau\_j$  of state $j$, **are i.i.d across trials** with pmf $p(\tau\_{j}|\boldsymbol{\theta}\_{d\,j})$ with subscript $j$ indexing the scale and shape parameters associated with stage $j$.
- The **"complete data" log-likelihood** for a single series $s$, given data $\mathbf{C}_s$ and the durations $\boldsymbol{\tau}_s$ of all $N$ states on trial $s$, is additive and equal to $log(p(\mathbf{C}_s\,\boldsymbol{\tau}_s | \boldsymbol{\theta})) = log\left(p(\boldsymbol{\tau}_s|\boldsymbol{\boldsymbol{\theta}}_d)\right) + log\left(p(\mathbf{C}_s\,\boldsymbol{\tau}_s | \boldsymbol{\theta}_b \right)$ where $\boldsymbol{\theta}_d$ are parameters of the stage duration distributions and $\boldsymbol{\theta}_b$ are parameters of the topographies. $log\left(p(\boldsymbol{\tau}_s|\boldsymbol{\boldsymbol{\theta}}_d)\right)$ is simply the log probability of stages having durations $\boldsymbol{\tau}_s$ while $log\left(p(\mathbf{C}_s\,\boldsymbol{\tau}_s | \boldsymbol{\theta}_b \right)$ is simply the ``gain`` computed for known stage durations and data.
- Durations of different states are assumed to be mutually independent as well. In consequence: $log\left(p(\boldsymbol{\tau}\_s|\boldsymbol{\boldsymbol{\theta}}\_d)\right)=\sum\_{j=1}^N log\left(p(\boldsymbol{\tau}\_{s\,j}|\boldsymbol{\boldsymbol{\theta}}\_{d\,j})\right)$.

Under these model assumptions, we have that

$$
\begin{equation}
\begin{split}
Q(\boldsymbol{\theta}|\boldsymbol{\theta}^{k})=& \mathbb{E}_{\boldsymbol{\tau}\ \sim\ p(. | \mathbf{C}, \boldsymbol{\theta}^k)}\left\lbrace log\left(p(\mathbf{C},\boldsymbol{\tau}|\boldsymbol{\boldsymbol{\theta}}) \right)\right\rbrace\\
& \mathbb{E}_{\boldsymbol{\tau}\ \sim\ p(. | \mathbf{C}, \boldsymbol{\theta}^k)}\left\lbrace log\left(p(\mathbf{C},\boldsymbol{\tau}|\boldsymbol{\boldsymbol{\theta}}) \right)\right\rbrace\\
& \mathbb{E}_{\boldsymbol{\tau}\ \sim\ p(. | \mathbf{C}, \boldsymbol{\theta}^k)}\left\lbrace \sum_{s=1}^S log\left(p(\mathbf{C}_s,\boldsymbol{\tau}_s|\boldsymbol{\boldsymbol{\theta}}) \right)\right\rbrace \\
& \sum_{s=1}^S \mathbb{E}_{\boldsymbol{\tau}_s\ \sim\ p(. | \mathbf{C}_s, \boldsymbol{\theta}^k)}\left\lbrace log\left(p(\mathbf{C}_s,\boldsymbol{\tau}_s|\boldsymbol{\boldsymbol{\theta}}) \right)\right\rbrace \\
& \sum_{s=1}^S \mathbb{E}_{\boldsymbol{\tau}_s\ \sim\ p(. | \mathbf{C}_s, \boldsymbol{\theta}^k)}\left\lbrace log\left(p(\boldsymbol{\tau}_s|\boldsymbol{\boldsymbol{\theta}}_d)\right) + log\left(p(\mathbf{C}_s\,\boldsymbol{\tau}_s | \boldsymbol{\theta}_b \right)\right\rbrace \\
& \sum_{s=1}^S \mathbb{E}_{\boldsymbol{\tau}_s\ \sim\ p(. | \mathbf{C}_s, \boldsymbol{\theta}^k)}\left\lbrace \left(\sum_{j=1}^N log\left(p(\boldsymbol{\tau}_{s\,j}|\boldsymbol{\boldsymbol{\theta}}_{d\,j})\right)\right) + log\left(p(\mathbf{C}_s\,\boldsymbol{\tau}_s | \boldsymbol{\theta}_b \right)\right\rbrace,
\end{split}
\end{equation}
$$

where $\boldsymbol{\tau}$ denotes the durations of all $N$ states for all $S$ series and $\boldsymbol{\tau}\ \sim\ p(. | \mathbf{C}, \boldsymbol{\theta}^k)$ is the joint posterior (which is just a nested product over the individual ones described earlier) for the latent variables $\boldsymbol{\tau}$. Lines 3 and 4 follow from assuming that different series are independent given the model (I.e., $\boldsymbol{\theta}$). Line 5 just spells out the complete data log-likelihood for series $s$. Line 6 further expands the complete data log-likelihood based on the assumption that the durations of different states are mutually independent as well.

Note, that when maximizing $Q(\boldsymbol{\theta}|\boldsymbol{\theta}^{k})$ with respect only to $\boldsymbol{\theta}_d$, $log\left(p(\mathbf{C}_s\,\boldsymbol{\tau}_s | \boldsymbol{\theta}_b \right)$ can be treated constant $M$ as it is not a function of the duration parameters. Thus,

$$
\begin{equation}
\begin{split}
Q(\boldsymbol{\theta}_d|\boldsymbol{\theta}^{k})=& \sum_{s=1}^S \mathbb{E}_{\boldsymbol{\tau}_s\ \sim\ p(. | \mathbf{C}_s, \boldsymbol{\theta}^k)}\left\lbrace \left(\sum_{j=1}^N log\left(p(\boldsymbol{\tau}_{s\,j}|\boldsymbol{\boldsymbol{\theta}}_{d\,j})\right)\right) + M\right\rbrace \\
& \sum_{s=1}^S \sum_{j=1}^N \mathbb{E}_{\tau_{s\,j}\ \sim\ p(. | \mathbf{C}_s, \boldsymbol{\theta}^k)}\left\lbrace log\left(p(\tau_{s\,j}|\boldsymbol{\boldsymbol{\theta}}_{d\,j})\right) + M\right\rbrace \\
& \sum_{s=1}^S \sum_{j=1}^N \mathbb{E}_{\tau_{s\,j}\ \sim\ p(. | \mathbf{C}_s, \boldsymbol{\theta}^k)}\left\lbrace log\left(p(\tau_{s\,j}|\boldsymbol{\boldsymbol{\theta}}_{d\,j})\right) \right\rbrace + M \\
& \sum_{s=1}^S \sum_{j=1}^N \sum_{d=1}^D\left\lbrace p(\tau_{s\,j}=d| \mathbf{C}_s\,\boldsymbol{\theta}^k)*log\left(p(\tau_{s\,j}=d|\boldsymbol{\boldsymbol{\theta}}_{d\,j})\right) \right\rbrace + M. \\
\end{split}
\end{equation}
$$

$Q(\boldsymbol{\theta}\_d|\boldsymbol{\theta}^{k})$ could just be maximized jointly for $\boldsymbol{\theta}\_d$ holding the scale and shape parameters from all $N$ states. However because the sum over the $N$ states will be maximized by maximizing each addend $j$ with respect to the duration parameters of that particular state, we can just maximize $Q\_{dj}(\boldsymbol{\theta}\_{d\,j}|\boldsymbol{\theta}^{k})$ separately for each state, where

$$
Q_{dj}(\boldsymbol{\theta}_{d\,j}|\boldsymbol{\theta}^{k})= \sum_{s=1}^S\sum_{d=1}^D\left\lbrace p(\tau_{s\,j}=d|\mathbf{C}_s, \boldsymbol{\theta}^k)*log\left(p(\tau_{s\,j}=d|\boldsymbol{\boldsymbol{\theta}}_{d\,j}) \right)\right\rbrace + M.
$$

Note that $Q\_{dj}(\boldsymbol{\theta}\_{d\,j}|\boldsymbol{\theta}^{k})$ could readily be computed in parallel for different states.

Regarding implementation: 

- The probabilities $p(\tau_{j\,s}|\mathbf{C}_s, \boldsymbol{\theta}^k)$ are estimated in the new ``estim_d_probs`` method. As described above, this is quite expensive currently. We could parallelize the function over series theoretically. 
- The evaluation and maximization of the $Q_{d\,j}$ functions happens in the new ``scale_shape_parameters`` method. Note that $M$ can be ignored for optimization purposes because, as mentioned, it is constant in the duration parameters. Hence it is not computed/used. Estimates from previous iterations are used as starting estimates for the optimization routine, so this should typically not be the bottle-neck. As mentioned above, we can speed this up further by including derivatives for the duration distributions with respect to their parameters or by parallelizing over states.

Now, **alternative optimization strategies** might work just as well - and might even be faster. For the old API @jelmerborst  has implemented an alternative strategy (see 356cc60dbeabaefe202c9f8e28607e6e28e1a9b1). It essentially computes the $Q_{d\,j}$ function above with the mean duration (the expectation there is taken over the posterior itself rather than the joint log likelihood), which is essentially like assigning the mean duration on each trial a probability of 1 in the full EM step. Also, instead of fully maximizing the resulting Q function only a single step is considered, which again introduces much less overhead and will still be feasible when the full EM step becomes too expensive. However, I do not see a guarantee that the resulting estimate will converge to the maximum likelihood estimate, so we cannot automatically guarantee consistency for example. So as always there will be a tradeoff, but I think having both would be great!

Cheers, have a nice weekend!

### DIFF
--- a/hmp/models/event.py
+++ b/hmp/models/event.py
@@ -751,8 +751,8 @@ class EventModel(BaseModel):
         eventprobs: np.ndarray,
         location: bool = True,
     ) -> np.ndarray:
-        """This method computes the probability of stage duration for each stage and each trial
-        given data and current parameters.
+        """This method computes the probability of peak-2-peak times for each of the n_events + 1
+        intervals and each trial given data and current parameters.
 
         Alternative output of the E-step of the EM algorithm.
 
@@ -775,41 +775,44 @@ class EventModel(BaseModel):
         """
         max_duration = np.max(durations)
         n_events = eventprobs.shape[-1]
-        n_stages = n_events + 1
+        n_intervals = n_events + 1
         n_trials = len(durations)
 
-        locations = np.zeros(n_stages, dtype=int)
+        locations = np.zeros(n_intervals, dtype=int)
         if location:
             locations[1:-1] = self.location
 
-        pmf_post = np.zeros([n_trials, max_duration, n_stages], dtype=np.float64)
+        pmf_post = np.zeros([n_trials, max_duration, n_intervals], dtype=np.float64)
 
-        # Compute p(tau_n = d | data, theta) where theta are current pars, i.e., the
-        # posterior over the latent states given data and current parameters. We need
-        # this do define the Q-function to be maximized during the M step with respect to
-        # scale and shape parameters of stage sojourn distributions.
-        for stage in range(n_stages):
+        # Compute p(tau_n = d | data, theta) where theta are current pars and tau_n is the
+        # peak-2-peak time for interval n, i.e., the posterior over the latent states given data
+        # and current parameters. We need this do define the Q-function to be maximized during the
+        # M step with respect to scale and shape parameters of peak-2-peak time distributions.
+
+        # Loop over intervals
+        for interval in range(n_intervals):
 
             for trial, T in zip(range(len(durations)), durations):
 
                 # p(o_{n-1} = t | C, theta) or p(o_{1} = t | C, theta) for this trial
-                ponset = eventprobs[trial, :, (stage-1) if stage > 0 else stage]
+                # where o_{n-1} is peak time for event n-1
+                ponset = eventprobs[trial, :, (interval-1) if interval > 0 else interval]
 
                 # Deal with censoring. Stage duration cannot exceed RT in current
                 # implementation.
                 # This makes edge cases special:
-                # For first stage probability of stage duration d is just probability
-                # of subsequent event happening at t = d.
-                # For last stage probability of stage duration d is just probability of
-                # previous event happening at t = T - d (with T = RT for that trial)
+                # For first interval probability of interval duration d is just probability
+                # of first event happening at t = d.
+                # For last interval probability of interval duration d is just probability of
+                # last event happening at t = T - d (with T = RT for that trial)
 
-                if stage == 0:
-                    didx = np.arange(locations[stage], T)
-                    pmf_post[trial, didx, stage] = ponset[didx]
+                if interval == 0:
+                    didx = np.arange(T)
+                    pmf_post[trial, didx, interval] = ponset[didx]
 
-                elif stage == n_stages - 1:
-                    didx = np.arange(locations[stage], T)
-                    pmf_post[trial, didx, stage] = ponset[T - didx - 1]
+                elif interval == n_intervals - 1:
+                    didx = np.arange(T)
+                    pmf_post[trial, didx, interval] = ponset[T - didx - 1]
 
                 else:
                     # Intermediate case, need joint probability of
@@ -817,19 +820,19 @@ class EventModel(BaseModel):
                     #   p(tau_n = d| C, theta)
                     # p(o_{n-1} = t, o_{n} =  t + d | C, theta) =
                     #   p(o_{n-1} = t | C, theta) * p(o_{n} =  t + d | C, theta)
-                    ponset2 = eventprobs[trial, :, stage]
+                    ponset2 = eventprobs[trial, :, interval]
 
-                    # Loop over possible **offsets** of next event
-                    for end_idx in range(locations[stage], T):
+                    # Loop over possible peaks of next event
+                    for end_idx in range(T):
 
-                        # Deal with censoring up to RT for intermediate case.
+                        # Deal with censoring at stimulus onset.
                         didx = np.arange(end_idx, -1, -1)
 
-                        pmf_post[trial, didx, stage] += (ponset2[end_idx] *
-                                                         ponset[:end_idx+1])
+                        pmf_post[trial, didx, interval] += (ponset2[end_idx] *
+                                                            ponset[:end_idx+1])
 
                 # Normalize again to ensure numerically valid pmf
-                pmf_post[trial, :, stage] /= np.sum(pmf_post[trial, :, stage])
+                pmf_post[trial, :, interval] /= np.sum(pmf_post[trial, :, interval])
 
         return pmf_post
 
@@ -845,7 +848,7 @@ class EventModel(BaseModel):
 
         It computes the scale and shape parameters maximizing the Q-function of the EM step;
         the expectation of the complete log-likelihood taken with respect to the posterior
-        distribution of the latent variables (state durations) given data and current parameters.
+        distribution of the latent variables (peak-2-peak times) given data and current parameters.
 
         Parameters
         ----------
@@ -873,25 +876,26 @@ class EventModel(BaseModel):
         durations = trial_data.durations[subset_epochs]
         max_duration = np.max(durations)
         n_events = eventprobs.shape[-1]
-        n_stages = n_events + 1
+        n_intervals = n_events + 1
 
-        locations = np.zeros(n_stages, dtype=int)
+        locations = np.zeros(n_intervals, dtype=int)
         if location:
             locations[1:-1] = self.location
 
-        # Compute p(tau_n = d | data, theta) where theta are current pars, i.e., the
-        # posterior over the latent states given data and current parameters. We need
-        # this do define the Q-function to be maximized during the M step.
+        # Compute p(tau_n = d | data, theta) where theta are current pars and tau_n is the
+        # peak-2-peak time for interval n, i.e., the posterior over the latent states given data
+        # and current parameters. We need this do define the Q-function to be maximized during the
+        # M step with respect to scale and shape parameters of peak-2-peak time distributions.
         pmf_post = self.estim_d_probs(durations,
                                       eventprobs[subset_epochs, :, :],
                                       location=location
                                       )
 
-        # Now M step - can be performed separately per stage
+        # Now M step - can be performed separately per interval
         new_time_pars = time_pars.copy()
-        for stage in range(n_stages):
+        for interval in range(n_intervals):
 
-            # Define Q function for stage j
+            # Define Q function for interval j
             def Qtauj(theta: np.ndarray) -> float:
                 lshape = theta[0]
                 lscale = theta[1]
@@ -901,14 +905,12 @@ class EventModel(BaseModel):
                 with catch_warnings():
                     simplefilter("ignore")
 
-                    lp = np.log(np.concatenate(
-                            (
-                                np.repeat(1e-15, locations[stage]),
-                                self.distribution_pdf(np.exp(lshape),
-                                                      np.exp(lscale),
-                                                      locations[stage],
-                                                      max_duration)[locations[stage]:],
-                            )
+                    lp = np.log(
+                        (
+                            self.distribution_pdf(np.exp(lshape),
+                                                  np.exp(lscale),
+                                                  locations[interval],
+                                                  max_duration),
                         )
                     )
 
@@ -919,7 +921,7 @@ class EventModel(BaseModel):
                     for trial in range(len(durations)):
 
                         # Compute second sum over d.
-                        E += np.sum(pmf_post[trial, :, stage] * lp)
+                        E += np.sum(pmf_post[trial, :, interval] * lp)
 
                 # Return negative expectation for optim
                 return -E
@@ -927,7 +929,7 @@ class EventModel(BaseModel):
             # And maximize
             opt = scp.optimize.minimize(
                 Qtauj,
-                np.log(time_pars[stage, :]),
+                np.log(time_pars[interval, :]),
                 method="L-BFGS-B",
                 options={
                     "gtol": 0,
@@ -942,12 +944,12 @@ class EventModel(BaseModel):
             if not opt["success"]:
                 warn(
                     (
-                        "M step step for distribution parameters did not converge! "
-                        f"Message was: {opt['message']}"
+                        f"M step step for distribution parameters of interval {interval} "
+                        f"did not converge! Message was: {opt['message']}"
                     )
                 )
             else:
-                new_time_pars[stage, :] = np.exp(opt["x"])
+                new_time_pars[interval, :] = np.exp(opt["x"])
 
         return new_time_pars
 
@@ -1213,7 +1215,7 @@ class EventModel(BaseModel):
             A 1D array representing the probability mass function for the distribution
             with the given shape and scale parameters, normalized to sum to 1.
         """
-        offset = 1 if location == 0 else 0
+        offset = 1
         p = self.distribution.pdf(np.arange(max_duration) + offset, shape, scale=scale)
         p = p / np.sum(p)
         p[np.isnan(p)] = 0  # remove potential nans


JoKra1:
Looks like tests are failing due to the change to ``distribution_pdf``.. I really think this was a bug, but fixing it should technically not break anything else - just lead to slightly different estimates perhaps.. Well, this will require some more thought perhaps

GWeindel:
Hi Josh, this looks really good and I already want to try it, I'll give a thorough review next week. 

Regarding tests don't worry, most of them are equality tests that I implemented to flag when a change modified the likelihood which of course happens with your bug fix. Note also that I made a very small dataset to do tests quickly thus this can also lead to edge cases and weird behavior, I'll have to think how to have a better trade-off.

JoKra1:
No rush, take your time! :-)

I noticed that with my updates, 2 of the test case fails were due to an invalid llk (inf). With the changes to ``distribution_pdf`` (this had really nothing to do with the EM code) the scale parameter in the cumulative case would for example collapse to nearly/exactly zero, resulting in the ``pmf`` evaluation returning only zeros as well, which ultimately caused the likelihood computation to fail... To address that, I added a standard step-length control loop to the EM step, which ensures that new parameters still result in a valid likelihood and otherwise shrinks the step.

Now the tests only fail because as you mentioned the results are different due to the change to ``distribution_pdf`` 😬. I will leave it up to you to decide how to proceed!

Finally, I realised that I was not fully awake, when looking at the M step and made some errors. I edited the top post to show the correct derivations step by step and implemented the changes in the code as well.

JoKra1:
Latest commit deals with ``location`` parameters. I am not 100% sure about the actual implementation here. Technically, if ``eventprobs`` has already absorbed the ``location`` parameters, accounting for them again during the M step feels redundant. However, it should generally not be factually wrong to do so, because we do actually assume a minimum duration for stages if ``location`` is True. I will have to think about this. For now, ``location`` defaults to ``True``.

While implementing this, I noticed a couple of bugs in the current implementation - some new ones and some that have carried over from the original Matlab implementation. An example of the latter, relates to the way the ``pmf`` is computed for stage durations.

As a reminder, that a stage is defined as the duration of the preceding event (if there is one) and the **additional** time until the onset of the next event. Only the first stage's duration is fully determined by the stage duration distribution. All the others have a preceding event, and so a minimum duration of ``location`` (This is another reason why ``pattern.width >= pattern.location``, as discussed in #262). Thus ``pmf[location,stage] = self. self.distribution_pdf(1)/V`` where $V$ is a normalisation constant. Essentially, for all stages but the first ``self.distribution_pdf`` are defined for the time between two event onsets - the "duration" (i.e., ``location=width``) of the first of the two (there is another bug in the code related to this, because ``locations`` in the code is currently computed incorrectly - assuming no minimum duration for the last stage, which does however have a preceding event. I will track that in a separate issue).

So, ``self.distribution_pdf`` really only has support on the closed interval $(0,D-loc)$, while ``pmf[:, stage]`` has support on the closed interval $(0,D)$. Both need to be valid probability distributions, but **this is currently not the case**. The problem is with how $V$ is computed, the normalisation constant for ``self. self.distribution_pdf``.

This is the current implementation:

This was implemented as:

```
pmf[:, stage] = np.concatenate(
                (
                    np.repeat(1e-15, locations[stage]),
                    self.distribution_pdf(time_pars[stage, 0], time_pars[stage, 1], max_duration)[
                        locations[stage] :
                    ],
                )
            )
```

You will quickly notice that summing over the first dimension for any given ``stage`` does not equal in 1, and this is not just because ``np.repeat(1e-15, locations[stage])`` should really be ``np.repeat(0, locations[stage])``, i.e., exactly numerically zero. The problem is that ``self.distribution_pdf`` computes the normalisation constant $V$ over $(0,D)$ inside of the method instead of $(0, D-loc)$ but then slicing happens afterwards. So,

```
self.distribution_pdf(time_pars[stage, 0], time_pars[stage, 1], max_duration)
```

sums to 1 (but it shouldn't) and

```
self.distribution_pdf(time_pars[stage, 0], time_pars[stage, 1], max_duration)[locations[stage] :] 
```

does not. There is another critical bug here ``[locations[stage] :] `` slices away the probabilities for ``self.distribution_pdf(np.arange(locations[stage]) + 1`` when it really should slice away the probabilities for ``self.distribution_pdf(D - np.arange(locations[stage], D) + 1``. Again, that is because ``pmf[:, stage]`` can have duration up to $D$, but for all but the first stage ``self.distribution_pdf`` can only have a maximum duration of $D-loc $.

The fix is to compute

```
pmf[:, stage] = np.concatenate(
                (
                    np.repeat(0, locations[stage]),
                    self.distribution_pdf(time_pars[stage, 0],
                                          time_pars[stage, 1],
                                          max_duration - locations[stage]),
                )
            )
```

instead, which results in valid probability distributions for both stages and just the time from after event $j$ to next event $j+1$.

This was quite hard to spot and the bug has actually been present already in the original Matlab implementation, so it will be good to fix this. :-)

GWeindel:
Hey Josh,

Thanks a lot for the review of the code! Your comments and suggestions are a good basis for discussion but I think a few clarifications are necessary before we continue.

HMP is not the exact implementation of the model defined by Anderson et al 2016, thus the departure from their paper and code. Anderson et al. model stages using  a pseudo linear HsMM. HMP models trial recurrent multivariate patterns whose time distance can be described by time-distributions. For that reason 1) HMP estimates the peak instead of the onset of events and 2) implements less constraining LLK computation that you point out and 3) the notion of `stage` (mostly) disappeared from the documentation and paper. What you propose is indeed closer to the model by Anderson et al 2016 than to the 2024 paper  https://doi.org/10.1162/imag_a_00400

This difference is the basis of the bugs you describe:
- Regarding also #268  the time distributions don't model onsets of events but peak-to-peak time, therefore the definition of location you suggest is not correct, this however makes me realize that indeed there might better ways to account for it. But in any case the last time distribution before the end of the interval (i.e. response) does not need a location in practice. Adding it has the problem of forcing events that peak close to the response (e.g. LRP/CPP) to be pushed away from their maximum likelihood position 
- `np.repeat(1e-15, locations[stage])` is necessary see #187. While strictly speaking you're correct putting this to 0 leads to overly constrained models because some durations will necessary be short for which we might not want the model to be too punitive. This would then impose to filter out trials with short latencies whose definition is hard to have.
- The imputation of the location through the censoring of the PDF is desired again this is more a technical comment but adding the location as you suggest leads again to degenerate models through invalid LLK. This is also more forgiving if inappropriate time distributions are provided (see Appendix 3 of https://doi.org/10.1162/imag_a_00400)

Thus all in all I think what you describe is a perfectly valid implementation of Anderson et al.'s model but this is not necessary what is being sought after here as HMP is currently more a single-trial ERP detection tool than a proper stage model a la Anderson. I do however see a value in having this implementation, maybe we can think of a `stage.py` model? I also tag @lvmaanen and @jelmerborst  as they might be interested in taking part to the discussion.

On a side note it is better to do one PR for each bug fix otherwise it gets hard to track. At this point I only see "pmf for a duration of zero" as a genuine one but maybe you'll prove me wrong. It is also better to compare the outputs on simulated data (preferably realistic as produced by `simulations.py`). I can give a try on your implementation soon if you don't have the time before that.

JoKra1:
> the time distributions don't model onsets of events but peak-to-peak time, therefore the definition of location you suggest is not correct, this however makes me realize that indeed there might better ways to account for it. But in any case the last time distribution before the end of the interval (i.e. response) does not need a location in practice. Adding it has the problem of forcing events that peak close to the response (e.g. LRP/CPP) to be pushed away from their maximum likelihood position

> The imputation of the location through the censoring of the PDF is desired again this is more a technical comment but adding the location as you suggest leads again to degenerate models through invalid LLK. This is also more forgiving if inappropriate time distributions are provided (see Appendix 3 of https://doi.org/10.1162/imag_a_00400)

That makes sense, then location can indeed be computed as is and #268 can simply be closed.

> np.repeat(1e-15, locations[stage]) is necessary see https://github.com/GWeindel/hmp/pull/187. While strictly speaking you're correct putting this to 0 leads to overly constrained models because some durations will necessary be short for which we might not want the model to be too punitive. This would then impose to filter out trials with short latencies whose definition is hard to have.

That really should not be the case. ``1e-15`` can for all practical purposed be considered zero and should not lead to very different results. If it does, that has me believe that there might be a problem somewhere else. In any case, the normalisation still needs to be addressed for ``pdf`` to be a valid pmf though! Given your clarification, I think the slicing

```
self.distribution_pdf(time_pars[stage, 0], time_pars[stage, 1], max_duration)[locations[stage] :] 
```

is correct then in the sense that you want to take out the short durations rather than the long ones due to it being about peak-2-peak! But then you should still compute self.distribution_pdf over ``np.arange(locations[stage], max_durations)`` and then normalise over that.

> Thus all in all I think what you describe is a perfectly valid implementation of Anderson et al.'s model but this is not necessary what is being sought after here as HMP is currently more a single-trial ERP detection tool than a proper stage model a la Anderson. I do however see a value in having this implementation, maybe we can think of a stage.py model? 

I disagree, if you compute the ``pmf`` as described by you (after correctly normalising) the M-step will just work assuming peak-to-peak distances - the math is agnostic to that 😬. Everything else stays the same (since I did not actually implement the locations changes I proposed in #268). Just replace any reference to "stage" and "stage durations" with "peak 2 peak distance" and "peak 2 peak time" in my derivations and you get the same outcome. :-)

JoKra1:
Indeed, after reverting the changes to the pmf computation and the 0 to 1e-15 both the original M step and the proposed one work, and agree with the direct MLE optimisation!

@GWeindel you are a wizard, I stand corrected and tip my hat to you!

So basically everything back to how it was before except for the + 1 being added to ensure no duration of zero. This could even be done only when estimating scale and shape, since for fixed shape of 2 it works without this change. Let me know if you prefer that @GWeindel!

On my side I now get:

```
EM scale estimates = [[2.3932909  4.24195097]
 [4.37902282 2.88255133]
 [5.07482022 2.35118855]
 [1.86490585 5.56750838]]
EM -llk = -382.0116346434982
MLE scale estimates = [[2.39015263 4.25157081]
 [3.99206992 2.99025681]
 [4.58140798 2.48803688]
 [1.87128438 5.55791778]]
MLE -llk = -383.0234727685181
```

when estimating both scale & shape. Which is much closer. I do worry a bit that the llk is close and still the shapes are different. So maybe something is still not correct in how I compute the pmf. Because if I also remove the +1 (allow for durations of zero) I get:

```
EM scale estimates = [[2.         5.11895161]
 [3.6983466  3.13262147]
 [4.19820951 2.58984988]
 [2.         5.11895161]]
EM -llk = -380.34259942532447
MLE scale estimates = [[1.78726829 5.11742962]
 [3.3374528  3.26057205]
 [3.8336257  2.71149249]
 [1.38707836 6.69449245]]
MLE -llk = -384.3731552854874
```

which also looks approximately right. But then the M step fails often because of invalid probability values that happen when trying to compute the probability of a zero duration for a shape < 2. Which is why the first and last shape are not changes..

So still some bugs, but the general direction appears to be right.

 See for code below:

```
import numpy as np

import hmp
import scipy as scp
from hmp.models import EventModel
from hmp.patterns import HalfSine
from hmp.trialdata import TrialData
from tests.test_io import init_data_large

event_c, epoch_data, info, sfreq, n_events = init_data_large()

hmp_data = hmp.transformers.ProjPCA(epoch_data, n_comp=5)
event_properties = HalfSine.create_expected(sfreq=hmp_data.sfreq)
trial_data = TrialData.from_transformer(hmp_data, pattern=event_properties.template)
em_model = EventModel(event_properties, n_events=n_events)

em_model.fit(trial_data, verbose=False,estim_shape=True)
time_pars = em_model.time_pars

_, em_groups, _ = em_model.group_constructor(
                trial_data, em_model.grouping_dict
        )

em_lkh, xreventprobs = em_model._distribute_groups(
    trial_data, em_model.channel_pars, em_model.time_pars,
    em_model.channel_map, em_model.time_map, em_groups, True
)

print("EM scale estimates =",em_model.time_pars[0].reshape(-1,2))
print("EM -llk =",-em_lkh[0])

def direct_MLE(theta:np.ndarray):
    
    
    modelD = EventModel(event_properties, n_events=n_events)
    dTimepars = time_pars.copy()
    modelD.fixed_time_pars = [0,1,2,3]

    dTimepars[0] = np.exp(theta.reshape(-1,2))

    modelD.fit(trial_data, verbose=False,
               estim_shape=False,time_pars=[dTimepars])
    
    _, groups, _ = modelD.group_constructor(
                trial_data, modelD.grouping_dict
            )
    
    # location must be true, otherwise the optimizer will end up
    # maximizing the parameters based on assumming that location=False
    # which is not the case when calling em_model.fit(trial_data, verbose=False) 
    lkh, xreventprobs = modelD._distribute_groups(
        trial_data, modelD.channel_pars, dTimepars,
        modelD.channel_map, modelD.time_map, groups, True
    )

    return -lkh[0]

MLE1 = scp.optimize.minimize(
                direct_MLE,
                np.log(time_pars[0].flatten()),
                method="L-BFGS-B",
                options={
                    "gtol": 0,
                    "ftol": 1e-9,
                    "maxcor": 30,
                    "maxls": 100,
                    "maxfun": 5000,
                }
            )

print("MLE scale estimates =",np.exp(MLE1["x"]).reshape(-1,2))
print("MLE -llk =",MLE1["fun"])
```

JoKra1:
Ok, if I only add + 1 to the support if location == 0, it seems to work the best:

```
EM scale estimates = [[2.39571398 4.24295109]
 [3.70606531 3.13497173]
 [4.24693776 2.572475  ]
 [1.8673584  5.56171441]]
EM -llk = -384.1107083533607
MLE scale estimates = [[2.39413068 4.24200189]
 [3.56241177 3.05213975]
 [4.10187296 2.51915125]
 [1.86841042 5.56430367]]
MLE -llk = -385.0254177910275
```

Should of course be validated in other simulations against the MLE (note that validating agains true scales/shapes might be disappointing, independent of which optimiser to use, since estimating both accurately might require lots of data) :-)

JoKra1:
Hey @GWeindel,

I have made some adjustments based on our discussions regarding the change to peak-2-peak distances and that seems to have been the missing piece. Thanks for all your help and patience with me getting up to speed!!

Now it works exactly like the MLE and both seem to converge to a relatively pronounced maximum, which is a good sign for the likelihood stability in general. At least that is suggested by profile plots of the llk over different shape parameter values with all other parameters fixed. See below (blue line is profile llk evaluated at EM parameters, orange at MLE, solid vertical line is EM estimate, dashed vertical line is MLE).

Once #270 is merged I will merge devel back into this pr. For now I just quickly hacked in the distribution shifting change by simply always adding + 1.

I will verify the agreement with the direct MLE on the tutorial models as a next step!

image

Edit: Here also the actual estimates for all ``time_pars`` of both optimisers:

```
EM estimates = [[2.40717711 4.23964294]
 [3.75257516 3.17545986]
 [4.4572713  2.55179817]
 [1.87472793 5.54696753]]
EM -llk = -382.9448603541785
MLE estimates = [[2.40280163 4.23383435]
 [3.77623905 3.17810357]
 [4.4743222  2.5553503 ]
 [1.87466675 5.54739561]]
MLE -llk = -382.95977359643877
```

JoKra1:
Can be closed as per our last discussion @GWeindel. I will keep this alive in my fork in case anyone ever wants/needs to get back to this and for this repo will focus on cleaning up #273 instead!

PR #244, opened 2025-08-17
New Estimation interface Implementation

kiante-fernandez:
This PR refactors the HMP estimation system with a cleaner architecture. The changes improve code maintainability and make it easier to add new estimation methods in the future.

- Added `BaseEstimator` interface and `EMEstimator` implementation
- Created `EstimationResult` class for consistent output format
- Separated model logic from estimation algorithms

- Rewrote `fit()` method to use new framework
- Maintained backward compatibility - existing code still works
- Preserved all current functionality and parameters

- Removed old EM methods

- Existing functionality verified
- Backward compatibility confirmed

Example 
```python
from estimators import EMEstimator
estimator = EMEstimator(max_iter=1000)
result = model.fit(estimator=estimator)
```

PR #270, opened 2026-04-28
Shift distributions to handle durations of zero + step length control to ensure valid updates

JoKra1:
Hey @GWeindel,

This PR ensures valid* (*still don't sum to 1) pmfs. It does so by

1) Ensuring that peak-2-peak time distributions are treated as shifted whenever necessary to ensure that they do not return an infinite probability for a peak-2-peak distance of zero (e.g., possible when first peak happens at stimulus onset). The necessity to treat a distribution as shifted is determined case by case, with the Weibull requiring no shift.

2) Step 1 does not guarantee the pmf from collapsing to zero over the entire support for very small scale parameters. This would again cause a failure to fit. Thus, after each parameter update step-length control is performed, halving the parameter update vector until the log-likelihood becomes valid.

Because of these changes I have dropped the ``1e-15`` hack!

**Considerations:**
Technically, step 1 could be skipped for interval distributions with a ``location > 0``. After all, we take only the slice ``[locations[stage] :]`` when determining the actual pmf anyway. However, this would require an update to the normalisation. Currently, if an ``inf`` happens at the first index ``p = p / np.sum(p)`` will produce ``nan`` everywhere even if all other indices have valid probabilities. The pmf will then be set to zero everywhere due to ``p[np.isnan(p)] = 0``. I have previously advocated for normalising after the slice anyway (see #267, although the implementation there was incorrect resulting in wrong estimates when compared to the MLE) and I still think this is appropriate, because only then will

```
pmf[:, stage] = np.concatenate(
                (
                    np.repeat(0, locations[stage]),
                    self.distribution_pdf(time_pars[stage, 0], time_pars[stage, 1], max_duration)[
                        locations[stage] :
                    ],
                )
            )
```

actually be a valid distribution (sum to 1) for intervals with ``location > 0``. If we do this, then we could set the shift to zero for whenever ``location > pdf.shift``. The resulting implementation for ``distribution_pdf`` would then be:

```
shift = 0 if location >= self.distribution.shift else self.distribution.shift 
# Call below might produce inf at index 0 with shift == 0 for location != 0
p = self.distribution.pdf(np.arange(max_duration) + shift, shape, scale=scale)
# So slice here to exclude it before normalising
# Correctly drop short durations not long ones as in #267
p = p[locations[stage] :]
# Safe to normalise now
p = p / np.sum(p)
```

Let me know if you would like to include this in this PR or in a later one.


JoKra1:
Test fails are to be expected since the shift of distributions will change the llk of the final model. Let me know if you want me to fix them!

JoKra1:
Something to think about: given the theoretical possibility of interval times of zero I would suggest making the ``Weibull`` the default distribution for HMP models. It is the only one that naturally and theoretically handles this possibility.

hmp/models/event.py
```diff
@@ -530,6 +530,10 @@ def EM(  # noqa
             # As long as new run gives better likelihood, go on
             lkh_prev = lkh.copy()
 
+            # Storage for step-length control
```
GWeindel, line 542: Looks good but not sure how safe this is as the group and fixed parameters are not handled in the control loop

hmp/models/event.py
```diff
@@ -542,36 +546,57 @@ def EM(  # noqa
...
+                        new_time_pars[time_map[:, p] == p_set, p, :] = np.mean(
+                            new_time_pars[time_map[:, p] == p_set, p, :], axis=0
                         )
 
+            # Step length control to ensure parameter updates result in valid llk
+            for icor in range(31):
+
+                if icor == 30:  # just reset
+                    new_channel_pars = channel_pars
+                    new_time_pars = time_pars
```
GWeindel: In this case shouldn't there be some warning?

hmp/models/event.py
```diff
@@ -542,36 +546,57 @@ def EM(  # noqa
...
                     if p_set >= 0:
-                        time_pars[time_map[:, p] == p_set, p, :] = np.mean(
-                            time_pars[time_map[:, p] == p_set, p, :], axis=0
+                        new_time_pars[time_map[:, p] == p_set, p, :] = np.mean(
+                            new_time_pars[time_map[:, p] == p_set, p, :], axis=0
                         )
 
+            # Step length control to ensure parameter updates result in valid llk
+            for icor in range(31):
+
```
GWeindel: Any magic in 30? I think it's probably OK, just wondering where that comes from

GWeindel:
I'll take care of the tests, the only critical comment is on dealing with group/fixed parameters. The tests are passing (except the cumulative and the equlity on the final model but I'll deal with that) so I guess it's OK but I'm wondering whether there isn't some corner case in the tests... I should do some coverage checks I guess

GWeindel:
Regarding the Weibull I'll think about it but it's not the most flexible one in my simulations so far

hmp/models/event.py
JoKra1: Wait I am confused, the control loop happens after the group loop so how can that be? We now have:

```
new_channel_pars[cur_group, channel_map_group, :] = c_par
new_time_pars[cur_group, time_map_group, :] = t_par
```

collecting parameters for a specific group, right? Then after we finish the group loop we now compute:

```
new_channel_pars = (new_channel_pars + channel_pars)/2
new_time_pars = (new_time_pars + time_pars)/2
```

whenever a step fails - this could be due to the parameters for any group - to half the update vector (e.g., ``new_channel_pars-channel_pars``) and try again. If this is repeated over and over again ``new_channel_pars-channel_pars -> 0`` and thus the new parameters for each group in ``new_channel_pars`` are shrunk back to the previous group specific parameters in ``channel_pars``.

Same for fixed parameters, we compute:

```
new_channel_pars[cur_group, fixed_channel_pars, :] = initial_channel_pars[
                    cur_group, fixed_channel_pars, :
].copy()
new_time_pars[cur_group, fixed_time_pars, :] = initial_time_pars[
    cur_group, fixed_time_pars, :
].copy()
```

before the control loop and since these are the same values as in ``channel_pars`` and ``time_pars`` (as they are kept fixed) the update vector as defined above will anyway be zero for those.

Or am I missing something?

hmp/models/event.py
JoKra1: It's my go to number. If we cannot fix an update vector by halving it 30 times there is usually something really wrong and I do not want to wait any longer since each iteration requires a likelihood evaluation (expensive for big models). We can make this a parameter if you prefer and expose it to fit?

hmp/models/event.py
JoKra1: Ya good point! Will add one.

hmp/models/event.py
GWeindel: No you're right I was confused sorry

JoKra1: No problem, thanks for double checking me! Marking as resolved.

hmp/models/event.py
JoKra1: Done. Marking as resolved.

hmp/models/event.py
JoKra1: Latest commit exposes this all the way to fit. I briefly thought about only adding this to EM since eventual merger of #244 will allow other estimators, but most of them will need some kind of step length control any well so they can just r-use that same parameter.

Let me know if you're happy with this or if I missed something!

JoKra1:
```
Run ruff check hmp
PLR0913 Too many arguments in function definition (11 > 10)
  --> hmp/models/event.py:90:9
   |
88 |          )
89 |
90 |     def fit(  # noqa: PLR0912, PLR0915
   |         ^^^
91 |         self,
92 |         trial_data: TrialData,
   |
help: Consider adding `@typing.override` if changing the function signature would violate the Liskov Substitution Principle

Found 1 error.
Error: Process completed with exit code 1.
```

LOL ruff, don't tell me what to do haha. I think it's fine and would just propose to override, but if you have a better idea let me know!

JoKra1:
> I'll take care of the tests, the only critical comment is on dealing with group/fixed parameters. The tests are passing (except the cumulative and the equlity on the final model but I'll deal with that) so I guess it's OK but I'm wondering whether there isn't some corner case in the tests... I should do some coverage checks I guess

Ya we should definitely add tests. Maybe you can create a simulation with really really short latencies between two events. The scale should just approach zero during fitting but the new control should shrink it back in case it drops too low.

GWeindel:
> ```
> Run ruff check hmp
> PLR0913 Too many arguments in function definition (11 > 10)
>   --> hmp/models/event.py:90:9
>    |
> 88 |          )
> 89 |
> 90 |     def fit(  # noqa: PLR0912, PLR0915
>    |         ^^^
> 91 |         self,
> 92 |         trial_data: TrialData,
>    |
> help: Consider adding ``@typing.override`` if changing the function signature would violate the Liskov Substitution Principle
> 
> Found 1 error.
> Error: Process completed with exit code 1.
> ```
> 
> LOL ruff, don't tell me what to do haha. I think it's fine and would just propose to override, but if you have a better idea let me know!

We can just initialize it at model creation and let advanced users modify it after init if necessary?

JoKra1:
> > ```
> > Run ruff check hmp
> > PLR0913 Too many arguments in function definition (11 > 10)
> >   --> hmp/models/event.py:90:9
> >    |
> > 88 |          )
> > 89 |
> > 90 |     def fit(  # noqa: PLR0912, PLR0915
> >    |         ^^^
> > 91 |         self,
> > 92 |         trial_data: TrialData,
> >    |
> > help: Consider adding ``@typing.override`` if changing the function signature would violate the Liskov Substitution Principle
> > 
> > Found 1 error.
> > Error: Process completed with exit code 1.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > LOL ruff, don't tell me what to do haha. I think it's fine and would just propose to override, but if you have a better idea let me know!
> 
> We can just initialize it at model creation and let advanced users modify it after init if necessary?

That works well indeed, feel free to move this around when you fix the tests!

GWeindel:
The tests are almost ready, you fix indeed prevents fit failure with events that are very close together. But I see that contrary to the previous version we now have some 
```
[event.py:1003]: RuntimeWarning: invalid value encountered in divide
  p = p /np.sum(p)
[event.py:862]: RuntimeWarning: invalid value encountered in divide
  eventprobs = eventprobs/eventprobs.sum(axis=0)
```
This is the case for the new tests (I'll push later today) but also in the data of tutorial 2 across several methods (random starting point, eliminative).

JoKra1:
> The tests are almost ready, you fix indeed prevents fit failure with events that are very close together. But I see that contrary to the previous version we now have some RuntimeWarning: invalid value encountered in divide eventprobs = eventprobs /eventprobs.sum(axis=0)
> This is the case for the new tests (I'll push later today) but also in the data of tutorial 2 across several methods (random starting point, eliminative).

Edit: sorry formatting is horrible from mobile..

Ya, that is to be expected. As mentioned change 1 does not prevent the mass from collapsing to zero over the entire support range for very small scales. If that happens now, all probabilities will remain at zero which will cause 0 event probs everywhere as well. Thus a division by zero happens. This was previously prevented by the insertion of 1e-15, at least for intervals with a non-zero location, resulting in some event probs (incorrectly, due to interval durations considered impossible now having all the mass) being > 0. However, as mentioned that is not really appropriate/desirable.

The step length control should still guarantee a valid fit in those cases, right? I would assume that it just prevents the scale from dipping too low. So I would suggest to wrap this step in a warn.filter() (also the normalization in the distribution_pdf call) and then in the EM loop implement a counter how often invalid steps are corrected and if this is not zero when the EM terminates raise a single warning saying there were N invalid steps or something like that. What do you think?

JoKra1:
Also very nice that the tests actually cover this!

JoKra1:
> and then in the EM loop implement a counter how often invalid steps are corrected and if this is not zero when the EM terminates raise a single warning saying there were N invalid steps or something like that. What do you think?

Actually, can we also keep a trace of the Parameter arrays that result in invalid steps? Like we to for the llk at this moment. That will be super handy for spotting problems with the model later down the line. Because if a specific interval duration always approaches zero, we might want to consider trying a model without that interval by dropping 1 event.

JoKra1:
> > and then in the EM loop implement a counter how often invalid steps are corrected and if this is not zero when the EM terminates raise a single warning saying there were N invalid steps or something like that. What do you think?
> 
> Actually, can we also keep a trace of the Parameter arrays that result in invalid steps? Like we to for the llk at this moment. That will be super handy for spotting problems with the model later down the line. Because if a specific interval duration always approaches zero, we might want to consider trying a model without that interval by dropping 1 event.

Actually, actually.. if we are to keep a trace we should collect ``time_pars == new_time_pars`` after the control loop, because inside of it ``new_time_pars`` might just be ``nan``s  in case ``eventprobs = eventprobs/eventprobs.sum(axis=0)`` encounters a division by zero since the M step will then be based on an array of ``nan``s itself.

hmp/models/event.py
```diff
@@ -542,36 +560,63 @@ def EM(  # noqa
...
                 for p_set in np.unique(time_map[:, p]):
                     if p_set >= 0:
-                        time_pars[time_map[:, p] == p_set, p, :] = np.mean(
-                            time_pars[time_map[:, p] == p_set, p, :], axis=0
+                        new_time_pars[time_map[:, p] == p_set, p, :] = np.mean(
+                            new_time_pars[time_map[:, p] == p_set, p, :], axis=0
                         )
 
+            # Step length control to ensure parameter updates result in valid llk
+            for icor in range(n_cor + 1):
```
JoKra1, line 585: As per my latest comment, I realised we can actually save us the control loop if parameters actually become ``nan`` because then the halving will also create ``nan`` update vectors. So we can just do

```suggestion
            if np.any(np.isnan(new_time_pars):
              new_time_pars = time_pars
            if np.any(np.isnan(new_time_pars):
              new_channel_pars = channel_pars
            
            # Step length control
            for icor in range(n_cor + 1):
```

Which is much cheaper. We can also guard the control loop then so that it is only entered if none of the above two conditions were true.

JoKra1: It should still be entered if only one of them was true, since any of the two new sets of parameters not ``nan`` might still cause a neg inf llk.

JoKra1: Sorry, I am making suggestions before thinking things through - I need coffee.

Actually what we should do is set

```
eventprobs[np.isnan(eventprobs)] = 0
```

after calculating

```
eventprobs = eventprobs/eventprobs.sum(axis=0)
```

exactly like you currently do in ``distribution_pdf``. That should result in a scale estimate of 0, which is not ``nan`` but will cause a neg inf ``llk``, to be handled correctly by the current step length control loop. So this suggestion is not necessary!

GWeindel:
Quick update, we can't have 0 location if we shift the distribution as you suggest. This leads to the EM to be stuck very frequently with an event of 0 duration (e.g . cumulative, random starting points)....

JoKra1:
> Quick update, we can't have 0 location if we shift the distribution as you suggest. This leads to the EM to be stuck very frequently with an event of 0 duration (e.g . cumulative, random starting points)....

I guess that is a consequence of essentially saying that ``p(d=0) <- p(d=1)``. The difference can be huge:

```
support = np.arange(5)
p = scp.stats.gamma.pdf(support,2,scale=0.5)
p /= np.sum(p)
np.round(p,decimals=2)
# output: array([0.  , 0.75, 0.2 , 0.04, 0.01])
```

Suddenly, a duration of zero has a much larger probability (0.75) than if we were not to perform the shift (0). What if, instead of computing:
```
shift = self.distribution.shift
p = self.distribution.pdf(np.arange(max_duration) + shift, shape, scale=scale)
```

we instead compute:

```
shift = self.distribution.shift
p = self.distribution.pdf(np.arange(max_duration), shape, scale=scale)
p[:shift] = 0
```

This assigns a zero probability to a duration of zero for all distributions that do not handle durations of zero.

**Note:** This effectively always happened already, with ``p = scp.stats.gamma.pdf(support,2,scale=x)`` returning 0 at the first index for every x. So in principle for this distribution this was already always computed as if location=1 for the interval before the first and after the last event. So we technically did not have location=0 before, at least for the Gamma and all other distributions that need to be treated as shifted. With this change that will continue to be the case. Alternatively, you could just set the first and last index for locations to 1 - but this will not be necessary for the Weibull, so I prefer the change to ``distribution_pdf``.

JoKra1:
To be clear, this would still

> ensure that they do not return an infinite probability for a peak-2-peak distance of zero

independent of the scale/shape combination - at least as long as this happens before the normalisation in ``distribution_pdf``. So this is still a valid solution ensuring a valid pmf, with the advantage that it should not cause the problems that you are experiencing since these should already have had a 0 at index 0.

GWeindel:
Hey Josh,

I made a PR to your branch : https://github.com/JoKra1/hsmm_mvpy/pull/2

Your fix did allow me to see that the cumulative method made unreasonable starting points, nice!

It looks all good for me so once you accept my PR on your branch, and the tests on this one are all passing, then we can merge in devel.

Great work!

JoKra1:
I saw, looks mostly good to me - I had one comment, which I think we should still address. I can fix this if you want, but you can also take a look yourself!

Glad that the cumulative method works again now, I am very happy we did not break it 😬!

GWeindel:
I probably won't have time before Friday but happy to help after that. What comment are you referring to?

JoKra1:
Tagged you, so I hope GitHub will point you there 😬

GWeindel:
Apparently not

JoKra1:
It should be in the new PR you made on my branch. It's a code comment so maybe that is why it's not notifying you for tags, but that would be odd..


Anyway, then we do this old school:
image

GWeindel:
Oups indeed I swapped it and adapted the tests

GWeindel:
There is a last point that we need to address. In further tests I realized that there can be situations where the cumulative method leads to invalid models that are no longer captured. That's rather a good thing but currently such fit are rejected (l.300 `events.py`):
```python

        if np.isneginf(lkhs.sum()):
            raise ValueError("Fit failed, inspect provided starting points")
```
Thus I think we should not raise an error but a warning so that in the case of the -ive methods we can continue the fit with fresh starting points.

I'll change my PR soon

JoKra1:
> There is a last point that we need to address. In further tests I realized that there can be situations where the cumulative method leads to invalid models that are no longer captured. That's rather a good thing but currently such fit are rejected (l.300 `events.py`):
> 
> ```python
>         if np.isneginf(lkhs.sum()):
>             raise ValueError("Fit failed, inspect provided starting points")
> ```
> 
> Thus I think we should not raise an error but a warning so that in the case of the -ive methods we can continue the fit with fresh starting points.
> 
> I'll change my PR soon

Good point about the warning! I also noticed that your last comment fix missed one change still, I marked that for you. :-) If you go back anyway, can you also address that?

JoKra1:
Tests looking good as well, so from my side this is ready for merge into devel! Thanks for all the help and discussions @GWeindel! 🎉

PR #271, opened 2026-04-30
fix tutorial 2, remove arguments from demo sim

GWeindel:


GWeindel:
@jelmerborst @JoKra1 fixed! I'll just merge to fix quickly but it's safe. I also rephrased some points and shortened a bit. I'll come back to it later

PR #277, opened 2026-05-11
Work on cumulative

GWeindel:
- fixes the computation of the number of maximum event (quickly but badly re-introduced after #272 
- Turns on the location for transform after discussions in PR #270 and issue #262  (@JoKra1)
- Splitted fastforward into sequential (i.e. should the time starting points be sequential to one-another) and fastforward (faster estimation if rejection of proposal). The default now takes more time but finds more events (@jelmerborst)
- Included the possibility of starting from a base fit (probably of interest for @rickdott) and overall structure closer to eliminative method

Note that the new default changes the fit in tutorial 3, maybe the generalization test for LLK improvement will fix that

### DIFF
--- a/hmp/models/cumulative.py
+++ b/hmp/models/cumulative.py
@@ -40,8 +40,8 @@ class CumulativeMethod(BaseModel):
         which likelihood increased regardless of whether a subsequent event was found.
         If False, Testing new starting point solely starting from the last detected event.
     fastforward : bool, optional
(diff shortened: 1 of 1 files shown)


GWeindel:
I'm merging this to avoid conflict with upcoming PR and will come back to the cumulative method for the generalization test we discussed (@jelmerborst @JoKra1)

PR #280, opened 2026-05-13
Default pattern

jelmerborst:
Main changes:
- default `HalfSine` template when building `TrialData`
- therefore, `event_properties` are now part of `TrialData`, rather than 'pattern' only. I think this makes sense, but it does change some of the calls along the way
- likewise, all models get a property `event_properties` instead of `pattern`, made all necessary changes for the different model classes

Questions:
- is it OK to make `event_properties` a property of `TrialData`? Does mean that code is not backward compatible. However, I think `event_properties` should be part of `TrialData`, and not of the model.
- shouldn't `location` be a property of the `models` instead of the `events`?
- should we rename `TrialData` to `PatternData`, or `CrosscorrelatedData`, perhaps?

Let me know what you think!


GWeindel:
This looks good to me, I'll fix the tests in the weekend to see if it all works and write my review at the same time

- I agree with `PatternData` I'll do the renaming
- We can make the `event_properties` of this `PatternData`, it makes indeed a lot of sense
- Regarding `location` I guess you're right but I'm not entirely sure, I'll see while fixing the rest

Don't worry about backward compatibility, we're in beta phase so we can break things, + the last beta is now the latest version so this will only be a problem when others switch to the newest beta or the 1.0.0

jelmerborst:
Thanks @GWeindel. Regarding `location`: I'm also not sure, but I'm leaning towards making it part of the `Models`, because a) it is not necessary to do the cross correlation, b) it is not actually a property of the defined event shape, and c) I can see the use case of trying different locations with the same cross-correlated data (or even changing location on the fly while estimating the models).

GWeindel:
Yes you convinced me! It's true that it's a property of the model as the censoring is done on the time distribution

hmp/models/base.py
```diff
@@ -38,35 +33,30 @@ class BaseModel(ABC):
 
     def __init__(
```
JoKra1, line 24: Constructor does not define an initial value for ``self.event_properties``, this is only set when calling ``fit``. However, the ``__getattribute__`` method does not actually check for existence and just points to ``self.event_properties``. 

Not sure whether that is a problem @GWeindel, but I don't think the linter will like this?

hmp/models/cumulative.py
```diff
@@ -104,6 +104,9 @@ def fit(
         -------
         None
         """
+
+        self.event_properties = trial_data.event_properties
```
JoKra1: > However, I think event_properties should be part of TrialData, and not of the model.

But here it becomes part of the model? I totally agree with your statement, also because having it as part of both introduces redundancies.

hmp/models/event.py
```diff
@@ -144,6 +142,15 @@ def fit(  # noqa: PLR0912, PLR0915
         -------
         None
         """
+
+        self.event_properties = trial_data.event_properties
+        if self.n_events > 1 and self.event_properties.location < self.event_properties.width:
```
JoKra1: This kind of goes against the reason formulated by @GWeindel in #263 for having a warning in a place where it is only called once (to notify unaware users) but not repeatedly (to annoy users who willingly ignore this).

I think the point is still a good one, so we should think about this!

hmp/trialdata.py
```diff
@@ -84,14 +86,18 @@ def from_transformer(cls, transformed, pattern, dtype = np.float32):
                 durations = durations.assign_coords({name: coord})
         data = data.unstack().stack(all_samples=['participant','epoch','sample']).\
             dropna(dim="all_samples")
+
+        if event_properties == None:
```
JoKra1: Checks for ``None`` should be handled via ``is None`` not ``==None`` (I think this is PEP 8 standard for Python because the latter can be ambiguous depending on whether you overwrite the equality operator of an object).

JoKra1:
>Thanks @GWeindel. Regarding location: I'm also not sure, but I'm leaning towards making it part of the Models, because a) it is not necessary to do the cross correlation, b) it is not actually a property of the defined event shape, and c) I can see the use case of trying different locations with the same cross-correlated data (or even changing location on the fly while estimating the models).

> Yes you convinced me! It's true that it's a property of the model as the censoring is done on the time distribution

Can we use this opportunity to unify what ``location`` and ``locations`` means across all methods? Currently, ``EventModel.location`` is an ``int``, but in ``estim_probs`` we compute:

```
locations = np.zeros(n_stages, dtype=int)
if location:
    locations[1:-1] = self.location
```

What if we just compute ``locations`` at init? Then all methods can access the same variable and we don't have to worry about it being computed differently in different places (#273 needs it in another place). This also opens up the opportunity to give more control to the user: for example, not everyone might be happy with zeroes at first and last index and they can then change ``locations`` after init in case they want a guaranteed minimum duration for first and last interval (thinking about @lvmaanen's comment during the last meeting). Also makes it easier to change ``locations`` during estimation as suggested by @jelmerborst, should we ever have the need.

hmp/models/base.py
jelmerborst: that was already the case in the previous version, so should be fine?

jelmerborst: In fact, I now removed it.

hmp/models/event.py
GWeindel: Yes but if we make location a property of the model then we've got no choice left

GWeindel:
> > Thanks @GWeindel. Regarding location: I'm also not sure, but I'm leaning towards making it part of the Models, because a) it is not necessary to do the cross correlation, b) it is not actually a property of the defined event shape, and c) I can see the use case of trying different locations with the same cross-correlated data (or even changing location on the fly while estimating the models).
> 
> > Yes you convinced me! It's true that it's a property of the model as the censoring is done on the time distribution
> 
> Can we use this opportunity to unify what `location` and `locations` means across all methods? Currently, `EventModel.location` is an `int`, but in `estim_probs` we compute:
> 
> ```
> locations = np.zeros(n_stages, dtype=int)
> if location:
>     locations[1:-1] = self.location
> ```
> 
> What if we just compute `locations` at init? Then all methods can access the same variable and we don't have to worry about it being computed differently in different places (#273 needs it in another place). This also opens up the opportunity to give more control to the user: for example, not everyone might be happy with zeroes at first and last index and they can then change `locations` after init in case they want a guaranteed minimum duration for first and last interval (thinking about @lvmaanen's comment during the last meeting). Also makes it easier to change `locations` during estimation as suggested by @jelmerborst, should we ever have the need.

Yes good idea, but should be a separate PR (yes I have other answers also :) )

GWeindel:
Working on this I'm starting to wonder whether this is a good approach.

Separating pattern from the model literally means that the pattern is not part of the model. A more practical consideration is that with this change one can fit a model on a given pattern, estimates parameters and transform on data cross-correlated with another pattern.

What you wanted is not having to pass event_properties twice, at cross-correlation and at model init. Maybe a better step would be to skip `TrialData` and pass the `transformed` data to a model after having initialized a model with the expected pattern. The cross-correlation then becomes part of the model fit/transform

jelmerborst:
I fully agree.  But why did you separate the cross correlation from model estimation to begin with? Was there a good reason for that?

GWeindel:
To avoid triggering a new crosscorrelation for each call to `EventModel` which would be annoying for parent methods (eliminative, cumulative). So to do this properly we need to make it cached, or make it implicit (e.g. define it from parent functions), or leave as is.

jelmerborst:
Then again, takes .5 a second for 5 pps on my slow laptop, so the overhead would be negligible. Seems like a fine solution for now to just move the cross correlation to trialdata.

GWeindel:
A regular dataset (~15000 trials) at 250Hz takes 7 sec on average. I agree this negligible but if you run a cumulative model with e.g. 10 folds and a mean RT of 500ms that could potentially give you 10 * 100 starting points so it add 2 hours for the fit. + the unnecessary CPU usage

jelmerborst:
Two options I see (but I could be wrong, I don't know the code that well yet):
1. in the CumulativeMethod create only one `EventModel`, which does the cross correlation on initialization. Then change `n_events` on every call but use the samel model for the rest. Same for eliminative.
2. allow EventModel to be initialized with either transformed data or cross correlated data. Easy enough to check.

GWeindel:
Yes but option 1 won't work as many things are initialised with n_events, so I think option 2 is nicer and addresses the problem of declaring event_properties twice. Do you want to do a new PR? Or modify that one?

jelmerborst:
I don't think that's true: initialisation only sets `model.n_events`, the `fit` method does the rest. Right?

Option 1 seems cleaner to me: shall I try that first and see if we can change the `CumulativeMethod` in a way that we only do the cross correlation once? If that doesn't work, I can always add the option to provide `TrialData` directly to `EventModel`.

One other question: shall I add transformed data directly to `BaseModel`? I can't really see a model class that doesn't need data, and it also solves the 'get_attributes' issue that @JoKra1 mentioned above.

I can adapt this PR.

GWeindel:
Yes indeed sorry we can do it after init.

Maybe it's worth a try but I would try on `eliminative` as `cumulative` will change quite much with #282 (or we first merge #282 that will yield less conflicts). We also have to account for the case where there are parallel fit as is the case in both parent methods.

What do you mean by adding transformed directly to `BaseModel`?

jelmerborst:
Picked option 2 in the end: allowing either transformed data or cross-correlated data as input to all model classes. Also added Pattern super class.

event model:
```
model = hmp.models.EventModel(n_events=3, data=preprocessed)
lkh, estimates = model.fit_transform(cpus=cpus)
```

cumulative:
```
model_cumulative = hmp.models.CumulativeMethod(data=preprocessed)
_, estimates_cumulative = model_cumulative.fit_transform(cpus=cpus)
```

eliminative:
```
model_elim = hmp.models.EliminativeMethod(data=preprocessed)
_, estimates_elim = model_elim.fit_transform(cpus=cpus)
```

hmp/models/base.py
```diff
@@ -6,75 +6,69 @@
...
-    location : int
-        Minimum duration between events in samples. Default is the event_width.
+    data : Data to fit the model on. One of two options:
+            1. data from BaseTransformer or xr.DataArray containing transformed data.
+            2. TrialData object.
+            In case of option 1, data is cross-correlated with the pattern in event_properties.
+            If event_properties is None, a half sine with 50 ms width is used.
+    event_properties : Pattern
+        The pattern and properties to use for cross-correlation. Default is
+        half sine with 50 ms width.
```
GWeindel: What about using `pattern` instead of `event_properties`? I suggested that at some point but it's long and not explictely related to the `patterns` module. Given that now there is no longer an instance of `pattern` compared to `event_properties` we could just consistently refer to `pattern`

hmp/models/base.py
```diff
@@ -6,75 +6,69 @@
...
+        data: Any,
+        event_properties: Pattern = None, 
         distribution: Any = None
     ):
-        self.pattern = pattern
+        
+        if isinstance(data, TrialData):
+            self.trial_data = data
+        else: #assume transformed (is checked later)
+            self.trial_data = TrialData.from_transformer(data, event_properties)
```
GWeindel: Sorry I did not anticipate that you wanted that change. This change (and the two changes to fit/transform) breaks the whole fit/transform architecture. A model needs to be initialized without data, it also needs to be able to be fitted using one dataset and transformed with another (see example in the addition of the crossvalidation module #282). IMO the fit/transform is a very useful new architecture  and also aligns well with sklearn for which we could use some methods due to this architecture.

An alterantive would be to give data to init in order to read the pattern info but not sure it's that elegant/explicit so I'm wondering whether there is a way out that still resolves the problem of declaring event_properties (or pattern) twice...

jelmerborst: Shame, I really liked my solution :). But I do see your point, so I'll move the data back to the fit methods. 

Two options again:
1. move `data` and `pattern` to `fit`.
2. use `pattern` at `init`, and `data` at `fit`. Data can either be `transformed` or `TrialData`, in the latter case I'll check whether `pattern` in the model corresponds to `pattern` in the `TrialData`, and the recommendation is to use transformed data from a user perspective.

I think option 2 works, well, because `pattern` is part of the `model`, while `data` is not. Also if you want to transform on other data, you do not want to change the pattern that was used to fit. (on a sidenote, later I would like to move `time_map` and `channel_map` also to `init`, as those also define the model (and it's cleaner for the LOOCV).

Do you agree with option 2?

jelmerborst: One additional question: OK if I move `compute_max_events` to `utils.py`. It doesn't work in the `BaseModel` and is anyway more of a util.

jelmerborst: And one more: what about setting `self.trial_data` in the `fit` and `transform` methods, and make `data` optional for `transform`? That makes it available for all submethods without having to pass it along, and you can check the data the model was fit on, while allowing to transform on other data.

GWeindel: - Yes I agree with option 2, that makes the most sense to me.
- For time and `channel_map` those are already part of init, do you mean transferring it to fit? That would make sense indeed
- For compute_max_event it works better to have it part of utils.py
- For trial_data as part of self I'm a bit reluctant to include data in any model object as this will bloat the object and make it hard to share, store, etc. (and may have some unexpected consequence on RAM usage that I may oversee). As you may have already seen if you use `fit_transform(trial_data)` the same data is used for both, is there another case where you think the data should be optional in `trial_data `?

jelmerborst: - `time_map` and `channel_map` are part of `fit` currently. I think they should be part of init, just like n_events, as they define the model. Anyway, issue for later when I fix the group models.
- I agree with your reluctance, I'm also afraid of bloating things. `Trial_data` is passed along quite a bit inside `fit` though, and I wonder if that also increase RAM usage (not sure if it does in fact). Given that `trial_data` cannot change during model estimation, what about adding it to `self` at the start of `fit`, and deleting it at the end?
- I see that I don't actually understand how `fit_transform` works :) `EventModel.transform() `does require data, but it doesn't get it...
- I would like to define `Patterns` without `sfreq`, so they are only instantiated when the data is added in `fit`, which will then call `create_expected` and add `event_width` and `sfreq`. The nice thing is that you can then use a default pattern in `init` without having to provide `sfreq`. Is that OK?

GWeindel: - Right the maps should be part of the model.
- For `trial_data` I don't undertsand why we need it in self, is it just to avoid the repeated data passing? We could add it to the object and remove it at the end but I do not see the benefit and there might be problems we don't anticipate
- What do you mean by "it doesn't get it"? In `EventModel.transform` it is a mandatory argument
- For `Patterns` without `sfreq` I indeed don't see a reason not to do it. To the contrary maybe that would be a first step to fit the model on e.g. downsampled data and transform on original sfreq! But then we need to adapt the `time_pars` based on `sfreq`

GWeindel:
So to sum up:
- [x] `event_properties` -> `pattern`
- [x] `TrialData` -> `PatternData`
- [ ] `time_map` and `channel_map` form `init` -> `fit`
- [x] move `compute_max_utils` `base` -> `utils`
- [x] `sfreq` as part of fit, `pattern` is expressed in milliseconds
- [ ] define Gamma distributions as function of sampling rate? (as to finish the possibility to fit on one `sfreq` and transform on another)

Anything else?
Usually most of these points would have been their own PR but let's just do it here, and while we are at it maybe it's better to do the location thing in #284 also here. When you are done I can branch on here and do what remains to be done + linting and testing

jelmerborst:
Moved pattern to `init` and data back to `fit`. For defining a `Pattern` you only need `event_width_ms`, the template itself is created later based on the data that is provided to `fit`. I changed the `Pattern` class quite a bit for that. I quite like how it is now, but curious to hear what you think.

Screenshot 2026-05-20 at 18 24 18

Pattern can be left out, `HalfSine` with 50 ms is the default. Data can be provided as preprocessed or already cross-correlated data.

- also implemented for `Cumulative` and `Eliminative`
- I ended up moving `compute_max_events()` to trialdata, as it requires `TrialData`.
- deleting trial_data at the end works fine, but I can also change it back to the old situation if your prefer, just means we need to pass it to almost all methods

The other things I can do later, but if you have time please check what I did so far.

jelmerborst:
I moved `location` to the model, and renamed `TrialData` to `PatternData`. Unless you want me to undo me making `pattern_data` an attribute of model (it is really quite neat now :)), I think it can be merged for now.

Moving `time_map` etc I will do later when I fix the grouping models, they still don't always work.

I can update the tutorials next week.

GWeindel:
OK I think this is final.
- all tests are now passing including linting
- changed the code so that sfreq is never used up to fit (or transform)
- implemented the location change of #284 here as it was just quicker this way
- pattern_data is never stored with the model object to avoid any unanticipated consequences

Overall I really like that!
For the tutorials I think we can wait, working on your PR also makes me want to define the time parameters in milliseconds rather than in samples so that we can fit a model on one samling frequency and transform on another (e.g. going through EM with 1000Hz data would be too slow but transform is just one pass in estim_probs so much faster!). 

@jelmerborst I suggest we wait to meet on Monday before merging?

GWeindel:
Fixed a bug I introduced in starting_points and that wasn't caught by the tests (but detected from now on)
I also updated the code of the tutorials but not really the text we need to go over it before the next release anyway

GWeindel:
I'm done with the tests. I'll do the locations tomorrow morning, if you want to merge already go ahead and I'll do a separate PR.

Very nice work :)

Issue #281, opened 2026-05-15
New crossvalidation module


GWeindel:
While we are at new modules 0:)

After @JoKra1's suggestion I'm working on adding a regularization to the `cumulative` method using a pseudo-kfold (pseudo in the sense that we do not recompute the PCA). This, along with the LOOCV that you plan @jelmerborst, we could put all of that into a `crossvalidation` module instead of `utils` or the `evaluation` discussed in #279 
I could be done by monday for a first draft.

PR #282, opened 2026-05-15
Added crossvalidation module and kfold to cumulative

GWeindel:
Added the crossvalidation module discussed in #281 (removed LOOCV.py)

In the cumulative method we now have the option to perform a (pseudo) k-fold validation strategy through the `kfold` argument. Seems to work quite well!

### DIFF
--- a/hmp/models/cumulative.py
+++ b/hmp/models/cumulative.py
@@ -16,9 +16,6 @@ try:
 except NameError:
     from tqdm import tqdm
 
-default_colors = ["cornflowerblue", "indianred", "orange", "darkblue", "darkgreen", "gold", "brown"]
-
-
 class CumulativeMethod(BaseModel):
     """Initialize the CumulativeMethod.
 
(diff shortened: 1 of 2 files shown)


GWeindel:
Adapted the acceptation rule by taking the median of the relative LLK improvement across the folds and comparing to tolerance. 
Not sure adding the tolerance is completely correct but it's intuitive that if we deem a LLK improvement in the EM to be too small to be considered, it should also be considered too small to add an event.

GWeindel:
@jelmerborst since you're into that currently we should make `kfold` a property of the fit rather than the model as I initially implemented here?

jelmerborst:
Yes I would say so. I would include everything that defines a model in `init` (n_events, grouping, maps, distribution) , and things that determine what happens during a call to fit only in `fit` (tolerance, max_iteration, cpus). I now note that we have quite some overlap in `init` and `fit` in `EventModel`, and also the distinction is not always clear. `kfold` is quite clear though.

GWeindel:
For me this is ready to merge. I think it's better to merge this one before #280 because it introduces less changes and thus less conflicts.

If ever @jelmerborst you want to merge this make sure you `squash and merge` using the drop down menu next to `merge pull request` otherwise the commit history will be too long

GWeindel:
Actually maybe @rickdott might have an opinion on that so I also request your review :)

rickdott:
Hi both,

I have nothing to bring against your theoretical considerations as to why cross validation is better during fitting. I tested the code on a small dataset and it seems similar. If this causes fewer noisy events I'm all for it! I would however like to see a when/why to use this in the tutorials perhaps.

Rick

Issue #283, opened 2026-05-17
Location when transforming data


GWeindel:
# Problem
Commit f59fc41 re-introduced location in transform which makes sense as we want the same LLK that we maximized. But this has the expected but annoying consequence of preventing single-trials with separated events but with a duration lower than location to be irrecoverable. For example in Tutorial 1 with the selected trial: 
```
for event in range(n_events):
    plt.plot(estimates_true.sel(trial=('sub-0',0), event=event), label=f"Event {event+1}",
            color=colors[event])
plt.ylabel('P(event)')
plt.xlabel('Time (sample)')
plt.xlim(0,225)
plt.legend()
plt.show()
```
Before Commit:
<img width="570" height="428" alt="Image" src="https://github.com/user-attachments/assets/1e3c35c6-8211-484c-8d36-ccb9ec684956" />

After commit:
<img width="473" height="341" alt="Image" src="https://github.com/user-attachments/assets/ac3c9490-5db8-4eba-bdd4-432e7ddcbf27" />

For that particular trial Event 3 is too close to Event 4, when transforming with location part of the weight gets shifted and make the recovery of the true time all of those events less good (see the single-trial time recovery plot when running with current devel compared to https://hmp.readthedocs.io/en/latest/notebooks/1-How_HMP_works.html)

The purpose of locations is to correctly estimate the parameters as written in #262
But once these parameters are correctly estimated the locations are not necessary if our goal is to estimate the location of the maximum probability event. 

# Solution

I don't see great solutions:
- one would be to add an argument to `transform` whether to turn on the location or not but the default is hard to choose as it depends on the intention.
- Alternatively we could estimate LLK with location and p(event) without but that makes two transform operations and might be confusing.

Thoughts @JoKra1 @jelmerborst ?

JoKra1:
> But this has the expected but annoying consequence of preventing single-trials with separated events but with a duration lower than location to be irrecoverable.

> For that particular trial Event 3 is too close to Event 4

So the simulation is "wrong"? As in, it simulates trials on which the true duration between two events is smaller than their width (resulting in overlapping patterns) - which we assume to be impossible when specifying location > width?

In that case my answer would be to change the simulation not the model code 😬. Because the model is performing as expected here, right?

[...]

PR #284, opened 2026-05-18
Moving locations

GWeindel:
As mentioned in issue #283 and in PR #280, this PR moves the definition of the locations as a property of `EventModel`.

Overall this makes a lot of sense as location is indeed part of the model and not the pattern.
This also unites the different definition in the previous code.

I updated tutorial 1 to show the effect of location as suggested in #283


GWeindel:
Still have to fix the tests though, will do later

GWeindel:
I'll pause that PR until PR #280 is merged

PR #287, opened 2026-06-03
Small fixes to new location implementation, only samples after init

jelmerborst:
Some small fixes, but mainly introduced EventModel.locations_samples to be used after init. That way we keep milliseconds for users but only samples in the rest of the functions. This for two reasons: a) less confusing in the long run, you can't forget to convert locations to samples in case we want to use them in more places than EM, and b) if we update locations during estimation in the future, we want to use samples.

### DIFF
--- a/hmp/models/base.py
+++ b/hmp/models/base.py
@@ -3,6 +3,8 @@ from abc import ABC, abstractmethod
 from typing import Any
 from warnings import warn
 
+import numpy as np
+
 from hmp.distributions import Gamma
 from hmp.patterndata import PatternData
 from hmp.patterns import HalfSine, Pattern
@@ -57,6 +59,35 @@ class BaseModel(ABC):
         else: #assume transformed (is checked later)
             pattern_data = PatternData.from_transformer(data, self.pattern)
         return pattern_data
+
+    def _time_to_samples(self, time, sfreq):
+        """Calculate samples (int) based on time(s).
+
+        Parameters
+        ----------
+        time : float | ndarray
+            Time or times that need to be converted to samples.
+        sfreq : sample frequency of data
+        """
+        return np.rint(time * sfreq / 1000).astype(int)
+
+    def _compute_max_events(self, pattern_data : PatternData, location : float):
+        """Compute max nr of events that fit in trial.
+
+        Parameters
+        ----------
+        pattern_data : Pattern
+            PatternData object
+        location : float
+            Location in ms.
+        """
+        min_dur = np.min(pattern_data.durations.values)
+        location_samples = self._time_to_samples(location, pattern_data.sfreq)
+        if self.pattern.width < location:
+            return int(np.floor((min_dur - self.pattern.width)/ \
+                            location_samples)) + 1
+        else:
+            return int(np.floor(min_dur / location_samples))
 
     @abstractmethod
     def fit(self):
--- a/hmp/models/cumulative.py
+++ b/hmp/models/cumulative.py
@@ -128,14 +128,15 @@ class CumulativeMethod(BaseModel):
         pattern_data = self._instantiate_data_pattern(data)
 
         end = pattern_data.durations.values.mean() if self.end is None else self.end
-        self.step = self.location*pattern_data.sfreq/1000 if self.step is None else self.step
+        if self.step is None:
+            self.step = self._time_to_samples(self.location, pattern_data.sfreq)
         if self.max_n_events is None:
-            max_n_events = int(np.floor(np.min(pattern_data.durations.values) /\
-                                        (self.location*pattern_data.sfreq/1000)))
+            max_n_events = self._compute_max_events(pattern_data, self.location)
         else:
             max_n_events = self.max_n_events
         #stop when not possible to insert event
-        end = int(np.rint((end - self.location*pattern_data.sfreq/1000 - 1)/self.step))
+        end = int(np.rint((end - self._time_to_samples(self.location,pattern_data.sfreq) \
+                - 1) / self.step))
 
         pbar = tqdm(total=end)  # progress bar
         n_events, j = 1, 1 # j = sample after last placed event
--- a/hmp/models/event.py
+++ b/hmp/models/event.py
@@ -173,7 +173,6 @@ class EventModel(BaseModel):
         None
         """
         pattern_data = self._instantiate_data_pattern(data)
-        self.locations_samples = np.rint(self.locations/1000 * pattern_data.sfreq).astype(int)
 
         # A dict containing all the info we want to keep, populated along the func
         infos_to_store = {}
@@ -257,7 +256,7 @@ class EventModel(BaseModel):
                         time_group = np.where(time_map[cur_group, :] >= 0)[0]
                         n_stage_group = len(time_group)
                         proposal_p[cur_group, time_group, :] = self.gen_random_stages(
-                            n_stage_group - 1)
+                            n_stage_group - 1, pattern_data.sfreq)
                         proposal_p[cur_group, fixed_time_pars, :] = initial_p[0, fixed_time_pars]
                     time_pars.append(proposal_p)
                 time_pars = np.array(time_pars)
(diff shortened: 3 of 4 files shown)


hmp/models/cumulative.py
```diff
@@ -69,7 +68,7 @@ class CumulativeMethod(BaseModel):
     def __init__(
         self,
         pattern: Pattern = None,
-        location: float = None,
+        location: int = None,
```
GWeindel: Given that it's millisconds we don't have to force int IMO

hmp/models/cumulative.py
```diff
@@ -132,7 +131,7 @@ def fit(# noqa: PLR0912, PLR0915
         self.step = self.location*pattern_data.sfreq/1000 if self.step is None else self.step
         if self.max_n_events is None:
             max_n_events = int(np.floor(np.min(pattern_data.durations.values) /\
-                                        (self.location*pattern_data.sfreq/1000))) + 1
+                                        (self.location*pattern_data.sfreq/1000)))
```
GWeindel: The `+1` is correct as first and last stage do not need a location

hmp/models/cumulative.py
```diff
@@ -158,8 +157,8 @@ def fit(# noqa: PLR0912, PLR0915
         # Iterative fit
         while j < end and n_events <= max_n_events:
             prev_j = j
-            event_model = EventModel(n_events=n_events, pattern=self.pattern,
-                                     tolerance=self.tolerance,distribution=self.distribution)
+            #event_model = EventModel(n_events=n_events, pattern=self.pattern, location=self.location,
```
GWeindel: Right that can be removed

hmp/models/cumulative.py
```diff
@@ -208,15 +207,14 @@ def fit(# noqa: PLR0912, PLR0915
...
-                                     location=self.location,
-                                     tolerance=self.tolerance, n_events=n_events)
+            event_model = EventModel(n_events=n_events, pattern=self.pattern, location=self.location,
+                                     tolerance=self.tolerance, distribution=self.distribution)
             event_model.fit(
                 pattern_data,
                 channel_pars=np.array([[channel_pars[:n_events, :]]]),
                 time_pars=np.array([[time_pars[: n_events + 1, :]]]),
                 verbose=False,
-                cpus=1,
```
GWeindel, line 219: The cpus argument is only needed for kfold, so here it can safely stay 1 (although cpus will not change behavior in this case)

hmp/models/event.py
```diff
@@ -103,22 +103,20 @@ def __init__(# noqa: PLR0913
         self.channel_map = np.zeros((1, self.n_events))
         self.n_cor = 30
 
-    def _set_locations(self, locations, pattern_width):
+    def _set_locations(self, location):
         """Set minimum distance between successive events."""
-        if locations is None:
+        if location is not None and isinstance(location,np.ndarray) and len(location) > 0: #array, must be array of length n_events + 1
+            assert len(location) == self.n_events + 1, "If location is np:array, should have length n_events + 1."
```
GWeindel: Good check indeed

hmp/models/event.py
```diff
@@ -103,22 +103,20 @@ def __init__(# noqa: PLR0913
...
+            self.locations = location
+        else:
             self.locations = np.zeros(self.n_events+1, dtype=int)
             if self.n_events > 1:
-                self.locations[1:-1] = pattern_width
-        else:
-            if isinstance(locations, int):
```
GWeindel, line 113: Ah I see why you enforce int earlier. Maybe we should allow floats at this point also and force int at conversion to samples but it doesn't really matter in practics

hmp/models/cumulative.py
jelmerborst: Gives divide by zero warnings in both `eliminative` and `cumulative`. Without the +1 it works without warnings.

hmp/models/cumulative.py
jelmerborst: why not use it here so we can parallel process for `EventModel`? Or is that only used with multiple starting points or multiple groups? In any case, if we can use multiple cpus in `EventModel` in the futture this'll make sure we use it.

hmp/models/event.py
jelmerborst: I actually think float is fine :)

hmp/models/event.py
```diff
@@ -173,6 +171,8 @@ def fit(  # noqa: PLR0912, PLR0915
         None
         """
         pattern_data = self._instantiate_data_pattern(data)
+        self.locations_samples = np.rint(self.locations/1000 * pattern_data.sfreq).astype(int)
```
GWeindel: It has to be done again in transform as it might be another sfreq. Thus it's not obvious to me why this is needed as it shows it's not a model attribute

jelmerborst: hmm tricky. Given that everything is in samples as soon as we fit (eg time pars, durations, template length) it makes more sense to me to have samples here. On the other hand, it is indeed not really a model attribute. Also, if we do something with locations while fitting, I think it's much easier to do that in samples than in milliseconds. Well, you decide :).

jelmerborst: as an in between solution: what about adding a method `_to_samples()` so we can use that to go samples when we need to? Or we do what we do with data, and that is not make it an attribute, but pass locations-as-sample to the functions. If you let me know what you prefer, I can implement it.

hmp/models/cumulative.py
GWeindel: I see this is because we need to take np.ceil not np.floor and then +1

hmp/models/cumulative.py
GWeindel: Right the future argument is a good reason

hmp/models/event.py
GWeindel: I don't see a good reason for why it needs to change but I might forget about some usages. Do you have a concrete example?

jelmerborst: for example on line 749 in `event.py` we used `self.locations` in `gen_random_stages()`. I now switched that to `self.locations_samples`. Either wrapping it in a function so we always calculate samples in the same way or passing it along could prevent such errors to some extent.

hmp/models/cumulative.py
jelmerborst: in fact it is wrong: we calculate how many events fit in the min RT, locations only come into play if they are longer than events. So I think it should be:

```
if location > width:
    max_events = (minRT - width) / location + 1   #because last event doesn't need location
else:
    max_events = minRT / location
```
Do you agree?

hmp/models/event.py
jelmerborst: also for instance in calculating max_events we need samples. I guess I just want to standardize the calcs to avoid mistakes. I'm fine with a `_to_samples()` function to do that, then we don't have to make it an attribute of the model.

GWeindel: Yes `_to_samples` makes sense it can then also be used to convert `time_pars` but we should then also have a `_to_ms` so that we can convert them back and forth. I can make a suggestion in another PR if you don't have the time

jelmerborst:
What do you think of this version @GWeindel ?

hmp/models/base.py
```diff
@@ -58,6 +60,35 @@ def _instantiate_data_pattern(self, data):
...
+        location : float
+            Location in ms.
+        """
+        min_dur = np.min(pattern_data.durations.values)
+        location_samples = self._time_to_samples(location, pattern_data.sfreq)
+        if self.pattern.width < location:
+            return int(np.floor((min_dur - self.pattern.width)/ \
+                            location_samples)) + 1
+        else:
+            return int(np.floor(min_dur / location_samples))
```
GWeindel, line 90:
```suggestion
        if self.pattern.width < location:
            return int(np.ceil((min_dur - self.pattern.width)/ \
                            location_samples)) 
        else:
            return int(np.ceil(min_dur / location_samples))
```

jelmerborst: I think we do need `floor`, as we can not fit extra events in. Say you have 10 samples as min duration, and a template of 3, 10/3 = 3.33, which should be 3. The +1 is also necessary in the first case, as I subtract `pattern.width` from the min duration first (to account for the last, location-less stage)..

PR #288, opened 2026-06-04
cleaned up event.py

jelmerborst:
Three changes:
- cleaned up event.py, removing attributes of the model that were also passed along + general cleaning
- added `cpus` argument to fit_transform in base
- added option to submit time_pars and channel_pars as 2D or 3D arrays, enabling:
```
model_3ev = hmp.models.EventModel(n_events=3)
lkh_3ev, estimates_3ev = model_3ev.fit_transform(data=preprocessed,cpus=cpus)

model_3ev_2 = hmp.models.EventModel(n_events=3)
lkh_3ev_2, estimates_3ev_2 = model_3ev_2.fit_transform(data=preprocessed,time_pars=model_3ev.time_pars, channel_pars=model_3ev.channel_pars, cpus=cpus)
```

Question: should we also return time_pars etc as 2D array if we only have one group? Or does that break code someplace else?

### DIFF
--- a/hmp/models/base.py
+++ b/hmp/models/base.py
@@ -99,4 +99,6 @@ class BaseModel(ABC):
 
     def fit_transform(self, data, *args, **kwargs):
         self.fit(data, *args, **kwargs)
-        return self.transform(data)
+
+        cpus = kwargs['cpus'] if 'cpus' in kwargs else 1
+        return self.transform(data, cpus=cpus)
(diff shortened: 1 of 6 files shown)


GWeindel:
Perfect! really needed. For the output I would however always keep the same dimensions irrespective of grouping. From what I understand it's bad practices to have outputs of varying number of dimension

PR #289, opened 2026-06-05
Updating tutorial 3

GWeindel:
- Updated the default `step` in cumulative, slightly slower but better exploration both in tutorial 3 and in other datasets
- Updated tutorial 3 with new #280 and a better introduction to the fit/transform architecture + some more comments and clarification and visual updates


jelmerborst (approved):
Looks all good.

PR #292, opened 2026-06-12
Group fixes

jelmerborst:
Group models were incorrect. `_distribute_groups` returned data in a different order than the original order, while code further down the line assumed the same order. This is fixed in the code below, as are some other things. I also improved plotting (sorry for doing that in the same PR...).

output

Ain't it pretty??

For some reason the tests result in a segmentation error on my computer, which is why I already created this pull request (though it might still fail one or more tests).

One question: `_distribute_groups` (now `_estimate_probs_groups`) returns a likelihood per group. Given that we optimize on the combined likelihood, I think that is what `EM` should return, or at least it should be returned by `transform(..)`. For now I included both the separate likelihoods as well as the combined likelihood in the estimates structure. What do you think @GWeindel ?

### DIFF
--- a/hmp/models/event.py
+++ b/hmp/models/event.py
@@ -118,7 +118,7 @@ class EventModel(BaseModel):
                 )
             assert n_events == (time_map.shape[-1] - 1) or n_events == channel_map.shape[-1], \
                 ("n_events, time_map and channel_map must indicate same number of max events")
-            
+
             #already add group names
             group_names = []
             group_mods = []


GWeindel:
Awesome! We should fix the tests for that, I broke the groups a while ago I think and it wasn't captured.
I'll take a closer look soon!

jelmerborst:
fixed :)

hmp/models/event.py
```diff
@@ -241,28 +262,39 @@ def fit(  # noqa: PLR0912, PLR0915
...
                     time_pars.append(proposal_p)
                 time_pars = np.array(time_pars)
-        elif time_pars.ndim == 3: #provided, doesn't include starting points
-            time_pars = np.array([time_pars])
-        elif time_pars.ndim == 2: #provided, doesn't include starting points and groups
-            time_pars = np.array([[time_pars]])
-        infos_to_store["starting_points"] = self.starting_points
-        infos_to_store["sp_time_pars"] = time_pars
+
+        elif time_pars.ndim < 4:
```
GWeindel, line 266: Looks good, I think we can separate the whole parameter formatting into different functions to lighten a bit that one. Can make a suggestion in a different PR

hmp/models/event.py
```diff
@@ -919,26 +939,28 @@ def _distribute_groups(
...
-                    "event": ("event", np.arange(self.n_events)[data_events]),
-                    "sample": ("sample", range(np.shape(likes_events_group[i][1])[1])),
-                },
-            )
-            xreventprobs = xreventprobs.assign_coords(trial_x_part)
-            xreventprobs = xreventprobs.assign_coords(group=("trial", groups[groups == cur_group],))
-            all_xreventprobs.append(xreventprobs)
-        all_xreventprobs = xr.concat(all_xreventprobs, dim="trial", join='outer')
+        # all_xreventprobs must have same order as pattern_data because
+        # subset_epochs is used later on eventprobs!
```
GWeindel, line 943: Right nice catch!

PR #294, opened 2026-06-18
IO refactoring

GWeindel:
I'm working on a complete refactoring of IO. The main ojective is to have better code but also to be able to read and preprocess any BIDS dataset.

So far this version works for BIDS (tests broken for now), not mne.raw nor mne.epochs yet.
It is quite easy to analyze a BIDS dataset, e.g. this is the only code I need to analyze ERP core flanker task after downloading it:

```python
import hmp

sfreq=250

# Which triggers to center the epochs on
centering_id = {
       "compatible/left" : 11,
	   "compatible/right" : 12,
	   "incompatible/left" : 21,
	   "incompatible/right" : 22,
}

# Which triggers indicate the response/the end of the duration to be modelled
response_id = {
	   "left/compatible/correct":111,
	   "left/incompatible/correct":121,
	   "right/compatible/correct":212,
	   "right/incompatible/correct":222
}

# BIDS argument, root and selection (datatype, session, task, ...)
bids_kwargs = dict(
    root = "../../data/Flanker", #Mandatory
    datatypes=['eeg'],#Mandatory
    tasks=['LRP'],#Mandatory
    subjects=['001','002'],
)

# Epoching arguments, anything that can be used by mne.Epoch
epoching_kwargs = dict(
    tmin = -0.2,
    tmax = 2,
)

# Preprocessing arguments, can have low_pass, high_pass, sfreq, etc.
preprocessing_kwargs = dict(
    sfreq = sfreq,
)

# Reading the data
epoch_data, info = hmp.io.read_bids_raw(
    bids_kwargs=bids_kwargs,
    epoching_kwargs=epoching_kwargs,
    preprocessing_kwargs=preprocessing_kwargs,
    centering_id = centering_id,
    response_id = response_id,
    # preprocessing_fn=preprocessing, # users can now also use their own preprocessing functions
    montage='biosemi64',
    cpus=-2, #Can also read data in parallel
)

hmp_data = hmp.preprocessors.ProjPCA(epoch_data, interval_id='rt',
                                    min_duration=0.2, # minimum RT of 200ms
                                    n_comp=10, #need for speed
)


model = hmp.models.CumulativeMethod()
lkh, estimates = model.fit_transform(hmp_data)
``` 

The only thing that changes for downstream functions is that the `participant` dimension has to be replaced by `recording`:

[...]

PR #295, opened 2026-06-23
Basedata

jelmerborst:
Implementation of BaseData, to replace `preprocessors` (currently still included). BaseData is a dataclass, typically using classmethods to construct the object from io data. It separates data cropping, rejecting epochs, projecting the data, and applying the variance operations.

**Multiple usage options:** 

Option 1 – very general classmethod:
`preprocessed = hmp.basedata.BaseData.from_io(io_data, interval_id='RT', crop=True, reject=True, apply_variance=True, projection_type='pca', projection_kwargs={'n_comp': 10})`

Option 2 – apply all operations:
`preprocessed = hmp.basedata.BaseData.from_io_all(io_data, interval_id='RT', projection_type='pca', n_comp=10)`

Option 3 – equivalent to ProjPCA (all operations with PCA)
`preprocessed = hmp.basedata.BaseData.from_io_all_pca(io_data, interval_id='RT', n_comp=.95)`

Option 4 – sequential application:
```
preprocessed = hmp.basedata.BaseData.from_io(io_data)
preprocessed.crop_epochs(interval_id='RT')
preprocessed.reject_epochs()
preprocessed.pca(n_comp=10)
preprocessed.apply_variance_ops()
```

**Later changes to object**

Object can also be changed later through one of the following methods:
```
center()
apply_variance_ops(whiten=True, common_variance=True, subject_zscore=True)
crop_epochs(interval_id = None, offset_start: float = 0, offset_end: float = 0, verbose=True)
reject_epochs(self, interval_id = None, min_duration: float = 0, max_duration: float = float('Inf'), 
                          reject_amplitude: float = None, verbose=True)
project(projection_type, weights = None, n_comp: float = None, center: bool = True,
               method_pca: str='svd', verbose=True)
pca(n_comp: float = None, center: bool = True, method_pca: str='svd', verbose=True)
pca_and_variance(n_comp: float = None, center: bool = True, method_pca: str='svd',
                         whiten=True, common_variance=True, subject_zscore=True, verbose=True)
apply_pca_weights(weights, center: bool = True)
```

Finally, I implemented `n_comp` so that it means the number of components retained if > 1, otherwise (0 < n_comp < 1) nr of components explaining at least n_comp% variance. If None, user input requested.

If this looks fine, I'll implement tests and remove preprocessors.

### DIFF
--- a/hmp/basedata.py
+++ b/hmp/basedata.py
@@ -456,7 +456,10 @@ class BaseData:
         """Get data from specified participants."""
         data = copy.deepcopy(data)
         data.data = data.data.unstack()
-        data.data = data.data.sel(participant=[participants], drop=False)
+        if isinstance(participants,list) or isinstance(participants,np.ndarray):
+            data.data = data.data.sel(participant=participants, drop=False)
+        else:
+            data.data = data.data.sel(participant=[participants], drop=False)
         data.data = data.data.stack(trial=['participant','epoch'])
         return data
 
--- a/tests/test_transformers.py
+++ b/tests/test_transformers.py
@@ -94,6 +94,15 @@ def test_proj_pca_custom_variants_newbasedata(init_data, n_comp, center, whiten,
     custom.reject_epochs()
     custom.project(projection_type='custom',center=center, weights=pca.weights)
     custom.apply_variance_ops(whiten=whiten)
+    data = hmp.basedata.BaseData.remove_participant(custom, 'a')
+    data = hmp.basedata.BaseData.get_participants(custom, ['a'])
+
+    #and pca pca based
+    pca = hmp.basedata.BaseData.from_io(epoch_data)
+    pca.crop_epochs()
+    pca.reject_epochs()
+    pca.pca_and_variance(n_comp=n_comp,center=center,
+                         whiten=whiten, method_pca='pca')
 
     assert custom.data.shape[1] == n_comp
     if whiten:


hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+import xarray as xr
+from scipy.linalg import eigh
+
+import hmp
+
+
+@dataclass
+class BaseData:
+    """
+    Based data classes containing all data necessary for estimating HMP models.
```
GWeindel:
```suggestion
    Base data classes containing all data necessary for estimating HMP models.
```

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+
+    Note that __init__ does not perform any of the operations!
+
+    Attributes
+    ----------
+    data : xr.DataArray with dimensions [sample, component, trial], coordinates that
+           describe the dataset including recording, subject, epoch, and a trial
+           MultiIndex, and attributes sfreq and offset. Typically obtained
+           through class method 'from_io(..)'.
```
GWeindel:
```suggestion
    data : xr.DataArray 
    	   Data with dimensions [sample, component, trial], coordinates that
           describe the dataset including recording, subject, epoch, and a trial
           MultiIndex, and attributes sfreq and offset. Typically obtained
           through class method 'from_io(..)'.
```

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+    weights : xr.DataArray with dimensions [channel, component] and coordinates
+            channel (Fp1, CPz, ..) and component (0, 1, ..).
+        Default = None
```
GWeindel:
```suggestion
    weights : xr.DataArray
    	    DataArray with dimensions [channel, component] and coordinates
            channel (Fp1, CPz, ..) and component (0, 1, ..).
            Default = None
```

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+    interval_id: str, optional
+        Name of the variable that contains the trial intervals in the epoch_data
+        used for cropping and rejection.
+        Default = 'rt'.
+
+    #Cropping parameters used
+    crop : bool, optional
+        Crop data based on parameters below.
+        Default = False
```
GWeindel:
```suggestion
        Crop data based on the durations and the parameters below.
        Default = False
```

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+    common_variance: bool = True
+    subject_zscore: bool = True
+
+
+
+    ##Classmethods
+
+    @classmethod
+    @typing.override
+    def from_io( # noqa: PLR0912
```
GWeindel: is the PLR0912 still necessary?

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+        base_data._format_io_data()
+
+        #if crop or reject, check for ms
+        if crop or reject:
+            rts_arr = base_data.data.coords[base_data.interval_id].values.copy()
+            max_rt = np.nanmax(rts_arr)
+            if np.nanmax(rts_arr) > 500:
+                warn_str = f"Found intervals with a max value value of {np.round(max_rt,2)}\n,\
+                            assuming intervals are in milliseconds and converting to seconds"
+                warn(warn_str)
```
GWeindel:
```suggestion
                warn(f"Found intervals with a max value value of {np.round(max_rt,2)}\n"
                            "assuming intervals are in milliseconds and converting to seconds")
```

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+
+        for i in range(len(self.data.data)):
+            if rts_arr[i] > 0:
+                # check amplitude
+                if (np.abs(self.data.values[i, :, :]) > reject_threshold).any():
+                    self.data.values[i, :, :] = np.nan
+                    rej += 1
+            else:
+                self.data.values[i, :, :] = np.nan
+
```
GWeindel, line 608: Your suggestions (also the decoupling with median centering) makes me realize we should to the Peak-to-peak method as in MNE (https://mne.tools/stable/generated/mne.Epochs.html)

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+        Default = None
+
+    #Projection parameters used
+    projection_type: str, optional
+        Type of projection to apply. 'pca' or 'custom' or None (identity)
+        Default = None
+    n_comp: int, optional
+        Nr of components retained if > 1, otherwise (0 < n_comp < 1) nr of components
+        explaining at least n_comp% variance are retained.  If None, user input requested.
+        Default = None
```
GWeindel, line 102: It would be more user friendly to have n_comp out of kwargs because they all have that (if Identity then n_comp = channels)

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+                for key, value in reject_kwargs.items():
+                    setattr(base_data, key, value)
+            base_data._reject_epochs(verbose=verbose)
+
+        #center
+        if projection_kwargs is not None and 'center' in projection_kwargs:
+            base_data.center = projection_kwargs['center']
+
+        if base_data.center:
+            base_data._center_data()
```
GWeindel, line 242: This might be slower and more RAM consuming that why it was in crop before but maybe with th enew default float32 that's now reasonable

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+
+
+    ## Public functions
+
+    def center_data(self):
+        """Center data per epoch, typically before projection."""
+        #warn of previous projection applied
+        if self.projection_type != 'identity':
+            warn("Projection applied previously, this operation might have\n \
+                unintended consequences.")
```
GWeindel, line 302: I think that's a bit the big problem here. Users can now do operations in any order and we might not want to make it that easy

hmp/basedata.py
```diff
@@ -0,0 +1,788 @@
...
+            input(
+                f"How many PCs would you like to retain?"
+                f"95 and 99% explained variance at components "
+                f"{np.where(np.cumsum(explained_variance_ratio) >= 0.95)[0][0] + 1} and "
+                f"{np.where(np.cumsum(explained_variance_ratio) >= 0.99)[0][0] + 1}; "
+                f"components till {np.where(explained_variance_ratio >= 0.01)[0][-1] + 1} "
+                f"explain at least 1% variance per component)?"
+            )
+        )
+        return n_comp
```
GWeindel, line 698: I removed that at some point because it is behaving a bit weirdly in some instances, plus the fact that we can now select on explained variance makes this not so useful?

GWeindel:
Maybe it would be interesting also to update tutorial 1 to show the steps, but then after merging #294 as it also changes the tutorials

hmp/basedata.py
jelmerborst: yes unfortunately.

hmp/basedata.py
jelmerborst: I can implement that later in a separate PR. Then we should think carefully which preprocessing steps are dependent on cropping (all other can be applied before in software of choice).

hmp/basedata.py
jelmerborst: is the case for the constructors that we will typically use.

hmp/basedata.py
jelmerborst: Couple of things:
- I will describe this in Tutorial 1
- users will typically use one of the predefined from_io_...() functions.
- I do think users should be aware of what they are doing, but that might be too optimistic :)

hmp/basedata.py
jelmerborst: I leave it for now, as I find it useful to demonstrate in workshops at least, and also to see for myself what is going on. Seems to work fine. If it doesn't at some point, please create an issue and I'll take a look at it.

jelmerborst:
Ready to go, I think. Removed preprocessors, adapted all files, fixed tests, added tests for the new options.

For now, I left `from_io_all_pca() `as it was, as I think it is clearer. I renamed `from_io` to` from_io_general`. Let's see what others say tomorrow.

hmp/basedata.py
```diff
@@ -0,0 +1,806 @@
...
+attributes are set correctly.
+
+Option 1, specify all information in the 'from_io' call:
+
+    preprocessed = hmp.basedata.BaseData.from_io(io_data, interval_id='RT', crop=True,
+                    reject=True, apply_variance=True, projection_type='pca',
+                    projection_kwargs={'n_comp': 10})
+
+Option 2, first build an object with from_io(io_data), then apply operations:
+
```
GWeindel:
```suggestion
Option 2, first build an object with from_io(io_data), then apply operations in the following order:

```

hmp/basedata.py
```diff
@@ -0,0 +1,806 @@
...
+    preprocessed.reject_epochs()
+    preprocessed.pca(n_comp=10)
+    preprocessed.apply_variance_ops()
+
+Includes methods to:
+    1. Crop the data from epoch time 0 (e.g. stimulus) up to a specified interval
+       (e.g. response), optionally including a fixed offset after the interval.
+    2. Reject epochs whose interval exceeds lower and upper interval limits
+       (`min_duration` and `max_duration`) or amplitude exceeds a threshold
+    3. Center the data (recommended for covariance based approaches)
```
GWeindel:
```suggestion
    3. Center the data
```
Also necessary for non-covariance based approaches

hmp/basedata.py
```diff
@@ -0,0 +1,806 @@
...
+    #Projection parameters used
+    projection_type: str, optional
+        Type of projection to apply. 'pca' or 'custom' or None (identity)
+        Default = None
+    n_comp: int, optional
+        Nr of components retained if > 1, otherwise (0 < n_comp < 1) nr of components
+        explaining at least n_comp% variance are retained.  If None, user input requested.
+        Default = None
+    center : bool, optional
+        Whether to center the data across the last dimension before projection
```
GWeindel:
```suggestion
        Whether to center the data within each epoch for each channel before projection
```

hmp/basedata.py
```diff
@@ -0,0 +1,806 @@
...
+        base_data = cls(data = epoch_data.copy(), weights=weights, interval_id=interval_id)
+
+        #process data from io
+        base_data._format_io_data()
+
+        #if crop or reject, check for ms
+        if crop or reject:
+            rts_arr = base_data.data.coords[base_data.interval_id].values.copy()
+            max_rt = np.nanmax(rts_arr)
+            if np.nanmax(rts_arr) > 500:
```
GWeindel:
```suggestion
            if max_rt > 500:
```

hmp/basedata.py
```diff
@@ -0,0 +1,806 @@
...
+        subject_zscore: bool, optional
+            z-score each component for each participant
+            Default = True
+        """
+        if whiten:
+            self.whiten = True
+        if common_variance:
+            self.common_variance = True
+        if subject_zscore:
+            self.subject_zscore = True
```
GWeindel: This might be error prone, if one gives `whiten = 0`. I think we can safely remove this and  a non-bool will throw an error
```suggestion
```

hmp/basedata.py
```diff
@@ -0,0 +1,806 @@
...
+        offset_start_samples = int(np.rint(self.offset_start * self.sfreq))
+
+        rts_arr[np.isnan(rts_arr)] = 0  # rejected during epoching or inexistant
+        rts_arr = np.rint(rts_arr * self.sfreq).astype(int)
+
+        #check nr of samples
+        min_rt = min(rts_arr[rts_arr > 0])
+        if min_rt < 10:
+           warn("The shortest interval is less than 10 samples. "
+                "Consider specifying too short trials using the `min_duration` parameter "
```
GWeindel:
```suggestion
                "Consider specifying rejecting too fast trials using the `min_duration` parameter "
```

hmp/basedata.py
```diff
@@ -0,0 +1,806 @@
...
+                "or increasing sampling frequency of the signal.")
+
+        time0 = np.argmin(np.abs(self.data.sample.values))
+        if time0 + offset_start_samples < 0:
+            raise ValueError("Offset before start is too large for the epoch data provided."
+                             f"Max is {time0/self.sfreq} seconds.")
```
GWeindel, line 540:
```suggestion
        time0 = np.argmin(np.abs(self.data.sample.values))
        min_sample = np.min(self.data.sample.values)
        if min_sample > offset_start_samples:
            raise ValueError("Offset before start is too large for the epoch data provided."
                             f"Max is {min_sample/self.sfreq} seconds.")
```
I made a mistake here previously

hmp/basedata.py
```diff
@@ -0,0 +1,806 @@
...
+        if self.max_duration is float('Inf'):
+            self.max_duration = (int(self.data.sample.max()) - self.offset_end\
+                                      + self.offset_start * self.sfreq)/self.sfreq
+        if self.min_duration == 0:
+            self.min_duration = 1 / self.sfreq
+
+        rts_arr = self.data.coords[self.interval_id].values.copy()
+        max_rt = np.nanmax(rts_arr)
+        if max_rt > 500:
+            rts_arr /= 1000
```
GWeindel: Given that this is repeated a lot it should be a function e.g. `_check_scale_ms`

hmp/basedata.py
jelmerborst: thanks, I couldn't figure out what it was supposed to do.

hmp/basedata.py
jelmerborst: Was wondering about it, but it's only repeated once currently, and the warning appears also only once.

jelmerborst: nm, fixed it after all

GWeindel:
Looks all good to me, it is slightly slower probably because of the _format and _unformat but it's marginal and RAM is fine

PR #296, opened 2026-07-06
Re-working basedata

GWeindel:
@jelmerborst I had to re-write quite a bit the basedata module.

1. Re-separated the different projection modules as this was not very modular and made the code quite long, now it's future proof if we want to add e.g. MCCA, etc.
2. Re-integrated crop and reject because they are inherently linked:	
	- separating centering doesn't work because we need the baseline for centering (otherwise short durations get quite weird)
	- a trial should be rejected if the a sample in the baseline exceeds `max_amplitude` as that means that the centering is off
	- More efficient
3. I made a general selection method in utils but we can't use it in basedata as that yields a circularity import error. Because you planned it for LOOCV it's better to put it in LOOCV with something like:

```python
from hmp.utils import coord_selection
    @staticmethod
    def remove_coord(data, value, variable):
        """Remove data from recording."""
        if isinstance(value, list):
            data =  coord_selection(data, value, variable, 
                                              method=lambda x, v: ~x.isin(v))
        else:
            data =  coord_selection(data, value, variable, np.not_equal)
        return data

    @staticmethod
    def get_coord(data, value, variable):
        """Get data from specified recordings."""
        if isinstance(value, list):
            data =  coord_selection(data, value, variable, 
                                              method=lambda x, v: x.isin(v))
        else:
            data =  coord_selection(data, value, variable, np.equal)
        return data

```

This changes quite much the code but we can still do something like:
```python
 hmp_data = hmp.basedata.from_io(io_data)
 hmp_data.crop_reject_epochs(duration_id='response_time')
 hmp_data.project(hmp.projectors.PCA(n_comp=10))
 hmp_data.apply_variance_ops()
```

Hope you agree and still like it :)

I haven't done the test linting yet because the idea is to first merge this in `refactor_io` and then update tests, tutorials etc. from there


jelmerborst:
> Re-separated the different projection modules as this was not very modular and made the code quite long, now it's future proof if we want to add e.g. MCCA, etc.

This is a nice change, thanks.

> Re-integrated crop and reject because they are inherently linked:
> separating centering doesn't work because we need the baseline for centering (otherwise short durations get quite weird)
> a trial should be rejected if the a sample in the baseline exceeds max_amplitude as that means that the centering is off

I disagree. It just means that rejection and centering should be done before cropping. I don't like this change, especially because I think you cannot _not_ crop now if you want to reject. I would like to split it again in two functions, but otherwise at least have the option not to crop. I think what we actually should do is crop from baseline until rt, reject, remove baseline. OK, that does mean it should be in one function, but just give me the option not to crop, please :)

> I made a general selection method in utils but we can't use it in basedata as that yields a circularity import error. Because you planned it for LOOCV it's better to put it in LOOCV with something like:

I also don't agree: I think all data classes should have data selection methods. There are enough situations in which you want to select part of the data. Why not just include it in here?

Other than that:
- you removed almost all attributes from init. I think this is a shame, because now you cannot just initialize your own BaseData (without going through the functions). Also, I thought it was useful to be able to use the attributes to see what was done to a dataset (though I guess now you can check whether they exist).
- all the oneliners are gone... which I guess is fine for me :).

GWeindel:
So do you typically just need an option to center? or do you also need an option to reject? In both case it should be relatively easy, but in the latter case we have to write carefully to distinguish the methods.

OK for the data selection. I can do so that any dataclass has it's own get/sel that wraps the selection function in utils. That would be best no? (I can do that for basedata in this PR and for patterndata in another PR)

For the attributes yes you can check whether they exist. I got confused by them as there would be four ways the attibutes would be set (`default()`, `from_init()`, just `BaseData()` and the call of each function). I think here it's clearer, at least to me, what is being set for each function.

For the onle-liners I can re-implement them if you need, as I don't know exactly what you need for LOOCV I left the `pca_and_variance` but `apply_pca_weights` is too PCA specific and can be written as `custom.project(hmp.projectors.Custom(weights=pca.projector.weights))` so still a one-liner! If there are others that are missing for you and that can't be done easily in one line we can re-intregrate them.

jelmerborst:
Yes, I would at least like to center before the projection, and maybe reject too short trials.

Data selection: if that's easiest sure, if it's quite short (as it was) then it can also be in the class itself. One thing: I made sure that the coordinates remain also if there's only one participant left. I think that's good behavior, so we can assume that participant (or subject or research part) is always present.

Attributes: OK.

You don't need to reimplement the one-liners. I do like `apply_pca_weights`, as that is something that we do relatively often. I don't really follow the code: 

`custom.project(hmp.projectors.Custom(weights=pca.projector.weights)`

shouln't it be

`base_data.project(hmp.projectors.Custom(weights=other_data.projector.weights)`?

GWeindel:
I implemented the `select_coord` method. That way you can select participant and even filter out short trials without going through the whole reject burder, see:
```python
a = hmp_data.select_coord(value='S2', variable='recording', method=np.equal)
not_a = hmp_data.select_coord(value='S2', variable='recording', method=np.not_equal)
a = hmp_data.select_coord(value=['S2'], variable='recording', method=lambda x, v: x.isin(v))
not_a = hmp_data.select_coord(value=['S2'], variable='recording', method=lambda x, v: ~x.isin(v))
long_trials = hmp_data.select_coord(value=1.5, variable='RT', method=np.greater)
```

And yes you are right it's `base_data.project(hmp.projectors.Custom(weights=other_data.projector.weights` I just used the example from the tests. 

For the coordinates preservation, I do not have a problem with the coordinates/dimension selection because I use `xr.DataArray.where()` and not `xr.DataArray.sel()` I think. So the dimensions should always be preserved, e.g. :
```python
long_trials = hmp_data.select_coord(value=1, variable='sample')
```
gives
```
BaseData(data=<xarray.DataArray 'data' (sample: 1, component: 10, trial: 900)>
```

for centering it's just `self.data -= self.data.median(['sample'],skipna=True)` as you did before, do you think we should still add a method to the risk of causing confusion with `center = True` in `crop_reject_epochs`? Because again it's not correct to do without the baseline in my opinion. HMP will work better for baseline corrected data than for data in which you center the whole trial without baseline.

If you agree with all that I can merge into `refactor_io` and then merge all that to `devel`

GWeindel:
OK this is all ready now, tests and tutorials included!

jelmerborst:
> I implemented the `select_coord` method. That way you can select participant and even filter out short trials without going through the whole reject burder, see:
> 
> ```python
> a = hmp_data.select_coord(value='S2', variable='recording', method=np.equal)
> not_a = hmp_data.select_coord(value='S2', variable='recording', method=np.not_equal)
> a = hmp_data.select_coord(value=['S2'], variable='recording', method=lambda x, v: x.isin(v))
> not_a = hmp_data.select_coord(value=['S2'], variable='recording', method=lambda x, v: ~x.isin(v))
> long_trials = hmp_data.select_coord(value=1.5, variable='RT', method=np.greater)
> ```

This is great, thanks! Especially including the coordinate preservation.

> for centering it's just `self.data -= self.data.median(['sample'],skipna=True)` as you did before, do you think we should still add a method to the risk of causing confusion with `center = True` in `crop_reject_epochs`? Because again it's not correct to do without the baseline in my opinion. HMP will work better for baseline corrected data than for data in which you center the whole trial without baseline.

No I don't think we need a separate method. I just would like the option to reject trials without cropping them. Ie I have a preprocessed and baselined dataset for which I want to remove short trials just for HMP. Wouldn't the correct order be:
1. crop to RT
2. reject based on length and/or amplitude
3. center
4. remove baseline
This way we could make them all optional. 

If you want, you can merge and I can implement this next week (if you agree with it).

GWeindel:
Yes OK that sounds good but then I would suggest to keep `crop_reject_epochs()` as is because it's optimised (we have to do a for loop for cropping and thus all other operations take less RAM than they would as extra vectorized steps). But we can then add `reject_epochs()`? Their respective function would be mostly separated because of this vectorization thing except maybe the reject on length?

jelmerborst:
I think we can implement all in `crop_reject_epochs()`. I suggest you merge for now, we finalize this and I finish LOOCV, add tests and update/add tutorials.

PR #300, opened 2026-07-18
Tutorials july26

jelmerborst:
Done!

Tutorial 4 is entirely new, the other are updated and some small things in the code needed fixing. I will now start working on Tutorial 5, but I'll add that to the LOOCV PR.


GWeindel:
Looks good! I like the ground truth new illustration, much better than the max_iteration = 0 I admit :)

For tutorial 3, the part 

```python
import os
cpus = os.cpu_count()
print('nr of cpus: ', cpus)

# We initialize a model without specifying a distribution or a pattern to demonstrate the use of base_data for modeling.
model = hmp.models.CumulativeMethod()
_, estimates_cumulative = model.fit_transform(base_data, cpus=cpus) #use all available cpus as this estimation can take a while
```
Is however not necessary as cumulative does not use the cpus except if you request kfold.

I also like the tutorial 4 so far (as it is not fully finished yet from what I see on the PR?) but could you add the fit/transform illustration also? I think it's important for the logic although I agree that it might not be the best use case.

P.S I see you do like contours now

GWeindel:
Oh can you also do the left/right thing in tutorial 3 but for the topo_timecourse?

jelmerborst:
> Looks good! I like the ground truth new illustration, much better than the max_iteration = 0 I admit :)

Glad you like it! Yes, I also like how it worked out this way.

> For tutorial 3, the part
> 
> ```python
> import os
> cpus = os.cpu_count()
> print('nr of cpus: ', cpus)
> 
> # We initialize a model without specifying a distribution or a pattern to demonstrate the use of base_data for modeling.
> model = hmp.models.CumulativeMethod()
> _, estimates_cumulative = model.fit_transform(base_data, cpus=cpus) #use all available cpus as this estimation can take a while
> ```
> 
> Is however not necessary as cumulative does not use the cpus except if you request kfold.

Haha, shows how well I know the code.

> I also like the tutorial 4 so far (as it is not fully finished yet from what I see on the PR?) but could you add the fit/transform illustration also? I think it's important for the logic although I agree that it might not be the best use case.

It does help to save before commit... I was on a train and a bit in a hurry. Now it's done.

What do you think of adding the fit/transform to tutorial 3 (as it is not about grouping)? Perhaps do a split half on the data to illustrate it?

> P.S I see you do like contours now

You live and learn... and I wanted to illustrate the lateralization better. :)

jelmerborst:
> Oh can you also do the left/right thing in tutorial 3 but for the topo_timecourse?

Not sure what you mean - left/right also for a whole topo_timecourse in Example 1? See also the now finished tutorial 4, where the last example is grouping for speed/acc/left/right.

GWeindel:
> > Oh can you also do the left/right thing in tutorial 3 but for the topo_timecourse?
> 
> Not sure what you mean - left/right also for a whole topo_timecourse in Example 1? See also the now finished tutorial 4, where the last example is grouping for speed/acc/left/right.

yes just to show that this also work for topo_timecourse

GWeindel:
> > Looks good! I like the ground truth new illustration, much better than the max_iteration = 0 I admit :)
> 
> Glad you like it! Yes, I also like how it worked out this way.
> 
> > For tutorial 3, the part
> > ```python
> > import os
> > cpus = os.cpu_count()
> > print('nr of cpus: ', cpus)
> > 
> > # We initialize a model without specifying a distribution or a pattern to demonstrate the use of base_data for modeling.
> > model = hmp.models.CumulativeMethod()
> > _, estimates_cumulative = model.fit_transform(base_data, cpus=cpus) #use all available cpus as this estimation can take a while
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is however not necessary as cumulative does not use the cpus except if you request kfold.
> 
> Haha, shows how well I know the code.
> 
> > I also like the tutorial 4 so far (as it is not fully finished yet from what I see on the PR?) but could you add the fit/transform illustration also? I think it's important for the logic although I agree that it might not be the best use case.
> 
> It does help to save before commit... I was on a train and a bit in a hurry. Now it's done.
> 
> What do you think of adding the fit/transform to tutorial 3 (as it is not about grouping)? Perhaps do a split half on the data to illustrate it?
> 
> > P.S I see you do like contours now
> 
> You live and learn... and I wanted to illustrate the lateralization better. :)

yes split half is a better idea, or just select participants to make the code easier?

GWeindel:
I checked tutorial 4 and I like it but there are a couple of minor issues:

- the link to the CB&B paper does not support the 3/4 as there there were only 2/3 events, so I suggest we replace with the imaging neuro where we report 3 and 4 with cumulative  [reported before](https://direct.mit.edu/imag/article/doi/10.1162/imag_a_00400/125469/Trial-by-trial-detection-of-cognitive-events-in)
- I would introduce the `cpus` only after independent model because here also, only grouped (or starting points) use parallel estimation.
- You wrote the LLK interpretation the other way around (maybe you though B/AIC?) the grouped model has a lower likelihood so it does not fit data better, not a problem as it has much less parameters but we might phrase it that way and also the other way around for 4 vs 2 (which is actually nicer).

I now understand what you meant with the LLK sums, I still think it's fine but if you find a nicer way of both keeping that info and avoiding sums without returning an extra output after `transform` (as that would be a bit much I think) I'm all in.

jelmerborst:
Good points! Not sure what I was thinking with the likelihoods... 

Could you still check if your like fit/transform in tutorial 3 this way?

GWeindel:
Oh wait actually in tutorial 4 the last point is still wrong regarding likelihood no? Also I don't understand why you get these filter warning, aren't you using the latest IO? I don't have it when running on `devel`

jelmerborst:
Good points:
- filter warnings, happens when code is run twice without restart
- turns out the llks differ considerably in different runs, I now added a warning (and am happy I'm not turning senile just yet)

GWeindel:
Damn that was a nasty bug that I introduced... fixed with #304

PR #303, opened 2026-07-20
Corrects behavior of computing max event when width < location

GWeindel:
As pointed out by @jiongjiongli in #283 there was a mismatch between sample and time domain when computing max number of events when `width < location`.
This is now fixed by just always taking `int(np.floor(min_dur / location_samples))` @jelmerborst I don't remember is there a rational for the previous branching of `if self.pattern.width < location:`? Because I don't undertand why this was necessary

### DIFF
--- a/hmp/models/base.py
+++ b/hmp/models/base.py
@@ -83,11 +83,7 @@ class BaseModel(ABC):
         """
         min_dur = np.min(pattern_data.durations.values)
         location_samples = self._time_to_samples(location, pattern_data.sfreq)
-        if self.pattern.width < location:
-            return int(np.floor((min_dur - self.pattern.width)/ \
-                            location_samples)) + 1
-        else:
-            return int(np.floor(min_dur / location_samples))
+        return int(np.floor(min_dur / location_samples))
 
     @abstractmethod
     def fit(self):


jelmerborst:
Yes the rational is that the last event doesn't have a location. So you really need `min_dur - pattern_width` to account for the last event, divide the rest by `max((pattern_width, location_samples))`, and then add 1 for the last event. I just don't think we need to branch, and can use:

```
           return int(np.floor((min_dur - self.pattern.width)/ \
                            np.max((location_samples, self.pattern_width))) + 1
```

GWeindel:
It's not only the last event that does not need a location. I'll close here because the required changes are done in #308 + I added an explanation in the PR as I see that this wasn't clear, also to me... (will also update the FAQ)

jelmerborst:
I'm indeed on vacation so won't have time to look at this in detail. However, I'm pretty sure the `max_events` calculation is wrong: it is only the last _event_ that does not have a location (the first stage also doesn't, but we only care about how many events we can place). We indeed always fit one event, which is why it should be 

`(t_min - event_width) / max(location_width, event_width) + 1`

1. (t_min - event_width): first subtract the width of the last event (the event without a location, an also be the first) from t_min
2. divide the leftover min_duration by the widest of location and event (location can be shorter than event)
3. plus 1 for the last event that was subtracted at step 1

GWeindel:
I laid it out in the other PR more clearly I hope, and the added a new test that shows that this computation is true as I can simulate 3 50ms events into 102ms (so 2 locations at 50 samples + the 2 samples for the distribution shif for a 1000Hz simulated signal).
image

The first and last event just stick to stimulus and response but as we discussed quite a while ago this is something we want to be able to happen (e.g. CPP) if the data suggest that

PR #308, opened 2026-07-22
Fix location and crosscorrelation

GWeindel:
Quite a heavy PR with lots of important fixes especially for those using different sampling frequency and event width but not only:

- The crosscorrelation is now done with padding from the data, the new parameter `offsets` (replacing `offset_end` and `offset_start`) in `hmp.basedata.BaseData` adds time before and after the start and end of durations. This is **major** as before that, the `same` mode for scipy was padding with zeros which depending on the data could be very bad. Now it uses the data prior stimulus and after duration end to pad! If no offsets are provided we output a warning and the behavior is the same as before (so not a good idea)
- ~~The template used for the cross-correlation is now adapted to exactly the sampling frequency instead of just subsampling from 1000Hz half-sine. This has some important consequences on the requested `width` for a `hmp.patters.HalfSine` object. Width should be choosen depending on the sampling rate (e.g. `52` ms if `50` ms desired for a sf of 250Hz). If they don't align the pattern is modified with a warning~~
- EDIT: the template is sampled by first picking an offset to make it symmetric if it's a half-sine
- The location was still wrong as we did not account for the shift in distributions implmented in #270  Now we explicitely subtract the shift of the distribution from the non-zero locations
- The max number of events is equal to the minimum duration divided by the location + 1 see below for details
- All tutorials were updated to this newest version

All this is validated by a new test that uses the simplest possible data source to model (see `test_general.py` that should never fail), should have done that a year ago sorry... All of these changes are breaking, the syntax only changed lightly but the HMP solutions will definitely not be the same (but for the best :)).
@jelmerborst I appoint you as a reviewer but I think you're on vacation? If so no worries we can merge and you'll check later when we draft a new release.
Other people who should be interested in that at least as a FYI: @lvmaanen @JoKra1 @rickdott 

Explication on location:
1) Recall that contrary to HsMM-MVPA, HMP models peak-to-peak times.
2) The location is implemented to avoid overlap of events during EM (but also in likelihood computation, see discussion #283)
3) Recently we have disconnected locations from event width so the location exists independently of an event width $W$ even though by default the location $L=W$
4) By this definition the minimum number $N$ of events with a location $L$ one can have given the minimum duration of $T_{min}$ is : $N = \frac{T_{min}}{L} + 1$
The $+1$ exists because by this definition any strictly positive $T_{min}$ there is at least an event that fits in the data as the location should not be enforced in the case where $N = 1$. if $N > 1$ the first peak does not need a minimum duration nor does the last one. Both can theoretically happen at the start and end of the durations (see the new test). So the formula above is the only correct computation assuming no shift in the distribution. 
**Importantly** in the case of a distribution with a shift (e.g. Gamma) the $T_{min}$ should be adjusted to the two location less stages (but with a shift) so stim to ev 1 and last ev to end so:
$N = \frac{T_{min}-2}{L} + 1$

### DIFF
--- a/hmp/basedata.py
+++ b/hmp/basedata.py
@@ -370,7 +370,7 @@ def from_io(epoch_data: xr.Dataset) -> BaseData:
     base_data.data.attrs["sfreq"] = epoch_data.sfreq
     return base_data
 
-def default( # noqa: PLR0913
+def default( # noqa: PLR0913, PLR0917
             epoch_data: xr.Dataset,
             duration_id: str = 'response_time',
             offsets: tuple | float = (0,0),
--- a/hmp/io/mne_raw.py
+++ b/hmp/io/mne_raw.py
@@ -18,7 +18,7 @@ from xarray import Dataset
 from hmp.io import preprocessing, utils
 
 
-def read_mne_raw(# noqa: PLR0913
+def read_mne_raw(# noqa: PLR0913, PLR0917
     recordings: list,
     centering_id: dict,
     event_id: dict = {},
--- a/hmp/models/event.py
+++ b/hmp/models/event.py
@@ -84,7 +84,7 @@ class EventModel(BaseModel):
         one of 'gamma','lognormal','wald', or 'weibull'
     """
 
-    def __init__(# noqa: PLR0913
+    def __init__(# noqa: PLR0913, PLR0917
         self,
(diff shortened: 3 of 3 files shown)


GWeindel:
I'll merge tomorrow morning after some last checks that are running tonight

PR #310, opened 2026-07-28
Changed operations in basedata

GWeindel:
The operations of subject/recording zscoring and common_variance are now False by default as the previous default does not generalize well so it seems quite data dependent (but overall both False are a better default so far). I also changed the order of these if they are True, and `recording_zscoring` is now not subtracting the mean (as that does not make sense if already centered, will mostly hurt) but instead just ensuring all PCs have the same variance across recordings.

I also changed the tutorials with the new recommendations + changed some statements in the notebooks 3 and 4, most important is probably the change in the grouping models @jelmerborst A group model cannot have a higher likelihood except it the EM algorithm finds a sub-optimal solution for one of the group. If all goes well, the case in Tutorial 4, the less parameters the lower the likelihood so I adapted the text.

Minor I also fixed the `pca_and_variance` shortcut.

I'll likely merge tomorrow morning after some further tests.


GWeindel:
Common variance operation is tricky as it increases likelihood (but compared to non-standardized data so doesn't make sense I think) but gives less reliable solution over several datasets. Thus I turned that off as at least it makes the data operations list shorter:

```
- epoching
- rejection
- median centering
- projection
- whitening 
```
As several projects depend on the PR I'll merge but might be good to come back to that later

PR #311, opened 2026-07-31
einsum speedup for calculating parameters

rickdott:
Change slow loop to use `np.einsum()`. On an older version this sped up `EventModel.fit()` by 1.5x on a toy dataset of ~1300 trials and similarly on a dataset of 17k trials. Can't test it on the new version so I am relying on automated testing through this PR :)

### DIFF
--- a/hmp/models/event.py
+++ b/hmp/models/event.py
@@ -664,23 +664,18 @@ class EventModel(BaseModel):
         time_pars : np.ndarray
             A 2D array of shape (n_stages, 2) with the estimated time parameters (shape and scale).
         """
-        channel_pars = np.zeros((eventprobs.shape[2], self.n_dims))
         # Channel contribution from Expectation, Eq 11 from 2024 paper
-        for event in range(eventprobs.shape[2]):
-            for comp in range(self.n_dims):
-                event_data = np.zeros((len(subset_epochs),
-                                       np.max(pattern_data.durations.values[subset_epochs])))
-                for trial_idx, trial in enumerate(subset_epochs):
-                    start, end = pattern_data.starts[trial], pattern_data.ends[trial]
-                    duration = end - start + 1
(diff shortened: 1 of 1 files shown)


GWeindel:
Looks nice, I'll try on a few other datasets but I think this is a good cheap speedup

rickdott:
Tested it further, here's what I did:
[hmp_speedtest_new_np.ipynb](https://github.com/user-attachments/files/30655376/hmp_speedtest_new_np.ipynb)

- Create minimum notebook on devel branch
- Update MNE (1.12.1), XArray (2026.7.0), Numpy (2.4.6) to rule out differences in package optimizations
- Test EventModel and CumulativeMethod running times by using cProfile, which shows per-function walltime use
- Compared devel and this branch (check the included notebook to see if I did anything wrong)

This is what I find:
image

Maybe when you ran it @GWeindel the changes were not actually in your notebook?

These were ~750 and ~5000 trial datasets, time savings seem to be similar regardless of set size.

GWeindel:
I gave this a bit more time, the speed up is indeed present.

Aligning the versions to yours, with the data in Tutorial 3 with a 4 event model, using `timeit` over 100 repetitions I observe in the `eventmodel_speedup`:
```python
8.63 s ± 56 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
vs. in current `devel`:
```
9.28 s ± 361 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
So it's a speed-up! That speed-up might also scale differently for the different parametrization and data but I'm confident it's going to be always positive.

(for the cProfile test I think you need to add the time spent in einsum and the time in get_...)

For git history I'll first merge this, then adapt in #244 before merging that one also

GWeindel:
The speed up is really appreciable in quite a lot of applications :1st_place_medal:

PR #322, opened 2026-08-03
fix: keep group index and result position apart in grouped estimation (#318)

### LINKED ISSUE #318: IndexError when a declared group has no trials
`EventModel._estim_probs_groups` collects per-group results by appending over `np.unique(groups)`:

```python
for cur_group in data_groups:
    likes_events_group.append(self.estim_probs(...))
```

but then indexes that list by group id:

```python
all_xreventprobs.data[np.ix_(groups == cur_group,
    range(likes_events_group[cur_group][1].shape[1]),
    self.channel_map[cur_group, :] >= 0)] = likes_events_group[cur_group][1]
```

If the group ids present in the data are not `0..n_groups-1` the two disagree, since the list is positional but the index is an id.

Reproduced with three declared groups (channel and time maps with 3 rows) over 4 trials:

- `groups = [0, 0, 1, 2]` works, likelihood -16.34
- `groups = [0, 0, 2, 2]`, so group 1 is declared but empty, raises `IndexError: list index out of range`

This can happen when a level of a grouping variable is absent after trial rejection, or when a participant subset does not contain every condition.

The fix is to index by position, for example `dict(zip(data_groups, likes_events_group))`, or to error early with a clear message if a declared group has no trials.

Noticed while working on #244, where the per-trial scatter I added inherits the same pattern and so raises a few lines earlier. May be of interest to @rickdott given #309.

MohammadYusif:
## Summary

- `EventModel._estim_probs_groups` collects the per-group results in a list built positionally
  over `np.unique(groups)`, but then indexes that list with the group itself. The two only
  coincide when every declared group occurs in the data — as soon as a group declared in
  `channel_map`/`time_map` has no trial left (a level absent after trial rejection, a
  participant subset missing a condition), event probabilities are written onto the wrong
  trials and the loop ends on `IndexError: list index out of range`.
- The per-group log-likelihoods had the same problem from the other direction: they were
  returned as a compacted array over only the groups present in the data, while
  `group_labels` covers every declared group, so `group_lkh` — and the `traces_group` /
  `xrtraces` built from it in `EM` — silently stopped lining up with the groups they describe.
- This keeps the position in the result list and the group index separate, and sizes the
  likelihood array on the declared groups, leaving `nan` for a group without trials.

As mentioned in the issue, the alternative would be to reject a declared group without
trials outright. I went with indexing by position because a model fitted on the full data
and then transformed on a subset that happens to miss a condition is a legitimate case, but
happy to switch if you'd rather not allow empty groups at all.

## Changes

- `hmp/models/event.py`: in `_estim_probs_groups`, iterate with
  `enumerate(data_groups)` so the result position and the group index stay apart;
  pre-size `likelihood` on the declared groups (`np.zeros(self.channel_map.shape[0]) * np.nan`)
  and fill it by group, so `group_lkh` stays aligned with `group_labels`; sum it with
  `np.nansum`, which is unchanged whenever every group is present.
- `tests/test_fixed.py`: `test_grouping_absent_group` fits the same trials twice with fully
  independent per-group parameters — once coded `0`/`2` with the declared group `1` left
  without trials, once coded `0`/`1` with only the two occurring groups declared — and
  asserts the likelihood and both parameter arrays match, that the group likelihoods are
  reported for every declared group, and that the empty one is `nan`.

Fixes #318

### DIFF
--- a/hmp/models/event.py
+++ b/hmp/models/event.py
@@ -947,7 +947,11 @@ class EventModel(BaseModel):
                     )
                 )
 
-        likelihood = np.array([x[0] for x in likes_events_group])
+        # likes_events_group is ordered by position in data_groups while the maps and
+        # group_labels are indexed by group. Both only coincide when every declared group
+        # occurs in the data, hence the two are kept apart below. Groups declared in the maps
+        # but without any trial (e.g. after trial rejection) keep a nan likelihood.
+        likelihood = np.zeros(self.channel_map.shape[0]) * np.nan
 
         # all_xreventprobs must have same order as pattern_data because
         # subset_epochs is used later on eventprobs!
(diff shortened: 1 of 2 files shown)


GWeindel:
Hi @MohammadYusif

Thank you for your pull request, could we maybe have a quick online meeting regarding your suggestion? You can send me an email at gabriel.weindel@unil.ch

PR #323, opened 2026-08-05
Paralellized the cpus>1, starting_points=1 situation (CumulativeModel)

rickdott:
Parallelized `estim_probs()` when no parallelization occurs elsewhere (`cpus >1`, `starting_points = 1`).

I think the current implementation would also work with `kfold != 1`, since that sets `cpus=1` on the `EventModel.fit()` call, which enters the serial execution.

This is however now the fourth 'level' of parallelization:
- folds
- starting points
- groups
- trials

Since all levels end up running `estim_probs()` I propose we unify parallelization to only run `estim_probs()` in parallel, and thus loop serially over folds, starting_points, and groups. Theoretically parallelizing at a higher level would be optimal, but practically I think it would often result in processes with unequal finishing times (due to different group sizes for example).

I tested 2, 4, 10, 20 participant datasets (~800 trials --> ~21400 trials) and 1, 2, 4, 8 cores:
image
Take the RAM results with a grain of salt, I don't think it's actually using less RAM, its counting the shared memory conservatively. Visual inspection of the system monitor showed that it was equal-ish.

### DIFF
--- a/hmp/estimators/em.py
+++ b/hmp/estimators/em.py
@@ -17,10 +17,14 @@ def _init_worker(pattern_data: PatternData, model):
     _WORKER_DATA["model"] = model
 
 
-def worker_estim_probs(channel_pars, time_pars, chunk):
+def worker_estim_probs(channel_pars, time_pars, chunk, max_duration):
     """Worker function to estimate probabilities of a chunk of trials."""
     return _WORKER_DATA["model"].estim_probs(
-        _WORKER_DATA["pattern_data"], channel_pars, time_pars, subset_epochs=chunk
+        _WORKER_DATA["pattern_data"],
+        channel_pars,
+        time_pars,
+        subset_epochs=chunk,
+        max_duration=max_duration,
     )
 
 class EMEstimator(BaseEstimator):
--- a/hmp/models/event.py
+++ b/hmp/models/event.py
@@ -667,6 +667,7 @@ class EventModel(BaseModel):
         channel_pars: np.ndarray,
         time_pars: np.ndarray,
         subset_epochs: list[int] | None = None,
+        max_duration: int | None = None,
     ) -> tuple[float, np.ndarray, np.ndarray]:
         """
         Estimate probabilities for events and compute the log-likelihood.
@@ -709,7 +710,8 @@ class EventModel(BaseModel):
         durations = ends - starts + 1
         cross_corr = np.vstack([pattern_data.cross_corr[s : e + 1] for s, e in zip(starts, ends)])
         dtype = pattern_data.cross_corr.dtype
-        max_duration = np.max(durations)
+        if max_duration is None:
+            max_duration = np.max(durations)
         gains = np.zeros((cross_corr.shape[0], n_events), dtype=dtype)
         for i in range(cross_corr.shape[1]):
             # computes the gains, i.e. congruence between the pattern shape
@@ -828,6 +830,7 @@ class EventModel(BaseModel):
 
         n_trials = len(subset_epochs)
         n_events = channel_pars.shape[0]
+        full_max_dur = int(np.max(pattern_data.durations.values[subset_epochs]))
 
         # split trials into disjoint chunks
         chunks = [c for c in np.array_split(subset_epochs, cpus) if len(c) > 0]
@@ -836,11 +839,17 @@ class EventModel(BaseModel):
         if pool is not None:
             results = pool.starmap(
                 worker_estim_probs,
-                [(channel_pars, time_pars, c) for c in chunks],
+                [(channel_pars, time_pars, c, full_max_dur) for c in chunks],
             )
         else:
             results = [
-                self.estim_probs(pattern_data, channel_pars, time_pars, subset_epochs=chunk)
+                self.estim_probs(
+                    pattern_data,
+                    channel_pars,
+                    time_pars,
+                    subset_epochs=chunk,
+                    max_duration=full_max_dur,
+                )
                 for chunk in chunks
             ]
 
@@ -848,9 +857,7 @@ class EventModel(BaseModel):
         # likelihood: sum over trials, so chunk log-likelihoods add exactly.
         likelihood = float(np.sum([r[0] for r in results]))
 
-        # eventprobs: each chunk is (chunk_trials, chunk_max_dur, n_events)
-        # max duration over the WHOLE subset, padding shorter chunks with zeros on the sample axis
-        full_max_dur = int(np.max(pattern_data.durations.values[subset_epochs]))
+        # eventprobs: each chunk is (chunk_trials, full_max_dur, n_events)
         eventprobs = np.zeros((n_trials, full_max_dur, n_events), dtype=results[0][1].dtype)
         trial_likelihood = np.empty(n_trials, dtype=results[0][2].dtype)
 


rickdott:
I think the module-level functions are ugly, but they do need to be instantiated/read before anything uses them. If sometime in the future we create workers in the `__init__()` function this needs to be above EventModel

GWeindel:
The speed-up is impressive and definitely needed, also RAM usage management is important as the previous parallelization attempt did not manage that very well. 

Because of that last point and the other estimation method that come in with their own parallelization (see #324) we need 1) to move this to the incoming `estimator` module (#244) so that this chunking sits for the EM estimator and 2) to scrap the other parallelization options to only retain that one that. While the former parallel loops had the advantage of being in higher level functions they, as you say, do not use the parallelization optimally because of waiting times. I also think this is more coherent from a user perspective (cpus > 1 always works the same).

How should we continue? I can help but only later in the month.

rickdott:
I'll need to get used to the new estimator anyway, so I can add it as a PR to be merged after #244 ? I'll unify the parallelization approaches.

GWeindel:
Awesome! The PR is merged so you can rebase/update

rickdott:
That was easier to get used to than I thought! Props to @kiante-fernandez (maybe you also want to have a look, since this now also touches your code).

I think I removed all other parallelization pathways.

GWeindel:
There is still parallelization for the eliminative and cumulative methods I wonder whether we should keep them or not. The data passed to workers is the same so there's no problems like in the group parallel loop, on the other end it might again make sense for consistency to parallelize always at the same level also.

Could you try timing eliminative with workers prior to `EventModel` call vs after? I can do later also of course.

kiante-fernandez:
> That was easier to get used to than I thought! Props to @kiante-fernandez (maybe you also want to have a look, since this now also touches your code).
> 
> I think I removed all other parallelization pathways.

I will take a look at this tomorrow. My biggest concern is understanding how the parallelization will play with JAX. Which we will need to prevent out of the box deadlocks as we prep to get the the pymc estimators running.

GWeindel:
> > That was easier to get used to than I thought! Props to @kiante-fernandez (maybe you also want to have a look, since this now also touches your code).
> > I think I removed all other parallelization pathways.
> 
> I will take a look at this tomorrow. My biggest concern is understanding how the parallelization will play with JAX. Which we will need to prevent out of the box deadlocks as we prep to get the the pymc estimators running.

I guess with sampling methods you can't parallelize that way so you can go for the chain level but that's not necessarily the best way to use large sets of cores. I think we need to have parallelization depending on the estimator?

rickdott:
I found the remaining parallelization over folds that I missed. Replacing it with within-fold parallelization is a bit inefficient as folds are a split over trials, and parallelization further splits over trials again. With normal usecases (high amounts of trials, relatively low amount of folds) it is still a very good speedup though (theoretically).

rickdott:
Regarding how it would work with JAX/pymc: I'm not familiar with that, but presumably we can have the estimator call EM with cpus=1 if necessary? This should always be a serial execution, which is useful if something up the chain (samplers?) is itself parallel

rickdott:
Oops, found a mistake that made the entire parallelization not work. I overreacted to the kfold change, it parallelizes fine even at small set sizes :)

rickdott:
Last commit speeds up PCA by ~1.5x, should grow linearly with trial count

hmp/models/event.py
```diff
@@ -141,26 +148,37 @@ def __init__(# noqa: PLR0913, PLR0917
         self.grouping_dict = grouping_dict
         self.time_map = np.zeros((1, self.n_events + 1)) if time_map is None else time_map
         self.channel_map = np.zeros((1, self.n_events)) if channel_map is None else channel_map
+        self._pool = None
```
kiante-fernandez: See other comments but holding the pool on the model is what forces `__getstate__`, and that changes pickling for every consumer of `EventModel`, not just the workers. There is also no matching `__setstate__`, so a worker's model ends up with no `_pool` attribute at all rather than None.

hmp/models/event.py
```diff
@@ -305,15 +327,23 @@ def fit(  # noqa: PLR0913, PLR0917
...
             estimator = EMEstimator(
                 tolerance=self.tolerance,
                 max_iteration=self.max_iteration,
                 min_iteration=self.min_iteration,
             )
 
-        result = estimator.fit(self, pattern_data, channel_pars, time_pars, groups, cpus)
+        try:
+            if cpus > 1:
+                self._pool = mp.Pool(processes=cpus, initializer=_init_worker, initargs=(pattern_data,))
```
kiante-fernandez: The pool is created for whatever estimator is passed, but only EM reaches `chunked_estim_probs. fit(..., estimator=MCMCEstimator(...), cpus=4)` forks four workers that are never used, and since importing the sampler pulls JAX in at module level, that is also the fork-after-JAX case JAX warns about. Creating it inside `EMEstimator.fit `instead would mean it only exists for the estimator that wants it.

hmp/models/event.py
```diff
@@ -305,15 +327,23 @@ def fit(  # noqa: PLR0913, PLR0917
...
+
```
kiante-fernandez: Also `mp.Pool` takes the platform default start method (?), so a fork on Linux and spawn on macOS and Windows. Might be worth pinning explicitly with `mp.get_context(...)`, ideally through one helper shared with the four other mp.Pool sites in hmp/io, so the choice is made in one place.

hmp/models/event.py
```diff
@@ -786,7 +811,82 @@ def estim_probs(
...
+        cpus : int
+            Number of trial-chunks to split into. cpus=1 reproduces the single call exactly
+            (one chunk == the whole subset).
+        pool : multiprocessing.Pool or None
+            If given and cpus > 1, chunks run via pool.starmap. If None, chunks run serially
+            in-process
+        """
+        # estim_probs accepts a boolean mask of full length; normalize to indices
+        # here so chunk slicing and reassembly both work on positional indices.
+        if subset_epochs is None:
```
kiante-fernandez, line 818: This builds a full-length index array, and `estim_probs` distinguishes a mask from indices by length, so it reads this back as a boolean mask and drops trial 0.  I think that means that no caller hits.
Passing `subset_epochs` straight through would avoid having two normalization rules for one concept. For ref `see tests/gen_data/generate_reference.py:86`

hmp/models/event.py
```diff
@@ -786,7 +811,82 @@ def estim_probs(
...
+
+        subset_epochs = np.asarray(subset_epochs)
+        if subset_epochs.dtype == bool:
+            subset_epochs = np.where(subset_epochs)[0]
+
+        n_trials = len(subset_epochs)
+        n_events = channel_pars.shape[0]
+
+        # split trials into disjoint chunks
+        chunks = [c for c in np.array_split(subset_epochs, cpus) if len(c) > 0]
```
kiante-fernandez, line 833: Optional.  I wonder if you can get even more speed on top of what you already have. `estim_probs` pads every trial to its chunk's maximum duration, so splitting in natural trial order leaves nearly every chunk padded to the global maximum. Sorting by duration before partitioning, and scattering results back by trial index rather than the running row counter below, might recover a decent part of the padding work (so you could test this).

hmp/models/event.py
```diff
@@ -786,7 +811,82 @@ def estim_probs(
...
+
+        # run estim_probs per chunk
+        if pool is not None:
+            results = pool.starmap(
+                _worker_estim_probs,
+                [(self, channel_pars, time_pars, c) for c in chunks],
```
kiante-fernandez: self is pickled once per chunk on every call. Now that the pool is long lived, the model could go into `_WORKER_DATA` alongside pattern_data in `_init_worker`, leaving only the parameter arrays and the chunk indices on the wire.

hmp/models/event.py
```diff
@@ -786,7 +811,82 @@ def estim_probs(
...
+            )
+        else:
+            results = [
```
kiante-fernandez, line 842: transform(cpus=4) never creates a pool, so self._pool is None here while cpus > 1, and we split the trials, run each chunk, then pad and reassemble, which is strictly more work than a single estim_probs call and no faster (?). The same applies to anything calling EMEstimator.fit directly rather than through EventModel.fit, which is what the sampler's EM seeding path does.

hmp/estimators/em.py
kiante-fernandez: this can go too since its only caller gets killed in the PR?

hmp/projectors/pca.py
```diff
@@ -121,21 +121,21 @@ def _pca(self, vcov_mat: DataArray, n_comp: int, channel: DataArray):
     @staticmethod
```
kiante-fernandez, line 120: touching more files that are speed related but not about parallelization per se. Is this easier to review as its own PR? But dropping the per-trial .sel() seems like a win.

hmp/models/event.py
GWeindel: Yes parallelization should be estimator specific

hmp/models/event.py
GWeindel: Definitely worth a try. If you can't do it now Rick we can open an issue on that later

hmp/projectors/pca.py
GWeindel: Should be separate PR but OK for this one

hmp/estimators/em.py
GWeindel: yes

hmp/models/event.py
rickdott: Nice idea, I think this only speeds up the serial path, and slows down the parallel path. For parallel we'd rather have all of the chunks be equal work. I tested it using this code:

```python
    def chunked_estim_probs(
        self,
        pattern_data,
        channel_pars,
        time_pars,
        subset_epochs=None,
        cpus=1,
        pool=None,
    ):
        """Chunk estim_probs across trial-chunks and reassemble.

        Returns [likelihood, eventprobs, trial_likelihood]

        Parameters
        ----------
        pattern_data, channel_pars, time_pars, subset_epochs :
            Passed through to estim_probs.
        cpus : int
            Number of trial-chunks to split into. cpus=1 reproduces the single call exactly
            (one chunk == the whole subset).
        pool : multiprocessing.Pool or None
            If given and cpus > 1, chunks run via pool.starmap. If None, chunks run serially
            in-process
        """
        if cpus <= 1:
            return self.estim_probs(
                pattern_data, channel_pars, time_pars, subset_epochs=subset_epochs
            )

        # Need indices in this case
        if subset_epochs is None:
            subset_epochs = np.arange(len(pattern_data.starts))
        subset_epochs = np.asarray(subset_epochs)
        if subset_epochs.dtype == bool:
            subset_epochs = np.where(subset_epochs)[0]

        n_trials = len(subset_epochs)
        n_events = channel_pars.shape[0]

        durations_subset = pattern_data.durations.values[subset_epochs]
        order = np.argsort(durations_subset, kind="stable")
        sorted_epochs = subset_epochs[order]

        # split trials into disjoint chunks
        chunks = [c for c in np.array_split(sorted_epochs, cpus) if len(c) > 0]
        pos_chunks = [c for c in np.array_split(order, cpus) if len(c) > 0]

        # run estim_probs per chunk
        if pool is not None:
            results = pool.starmap(
                worker_estim_probs,
                [(self, channel_pars, time_pars, c) for c in chunks],
            )
        else:
            results = [
                self.estim_probs(pattern_data, channel_pars, time_pars, subset_epochs=chunk)
                for chunk in chunks
            ]

        # reassemble
        # likelihood: sum over trials, so chunk log-likelihoods add exactly.
        likelihood = float(np.sum([r[0] for r in results]))

        # eventprobs: each chunk is (chunk_trials, chunk_max_dur, n_events)
        # max duration over the WHOLE subset, padding shorter chunks with zeros on the sample axis
        full_max_dur = int(np.max(pattern_data.durations.values[subset_epochs]))
        eventprobs = np.zeros((n_trials, full_max_dur, n_events), dtype=results[0][1].dtype)
        trial_likelihood = np.empty(n_trials, dtype=results[0][2].dtype)

        for (_, chunk_ep, chunk_tll), positions in zip(results, pos_chunks):
            c_dur = chunk_ep.shape[1]
            eventprobs[positions, :c_dur, :] = chunk_ep
            trial_likelihood[positions] = chunk_tll

        return [likelihood, eventprobs, trial_likelihood]
```
And found 10% slowdown when parallel, and ~5% speedup if serial. I don't think it is worth it to make an alternative code path (if cpus == 1) that triggers this.

rickdott:
Thanks for your feedback both! I considered and/or implemented all points.

**Sorting trials by duration speedup**
See the comment I added to Kiante's review comment

**multiprocessing context**
I am developing on ubuntu, which (with my versions of packages/Python) defaults to fork
(See https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods for more information)
I added this function to `utils.py`:
```python
def _get_mp_context():
    available_methods = mp.get_all_start_methods()
    for method in ["fork", "forkserver", "spawn"]:
        if method in available_methods:
            return mp.get_context(method)
```
Which is used everywhere mp.Pool was used.
With a small dataset, fork was much faster than spawn due to how spawn pickles/sends data to processes. Spawn did not profit from multiprocessing at all.

I think this is mostly some 'base' costs of spawn and that this will lessen with bigger data. I'll run a bigger test and comment on this PR when it is done. For now I think this is a sensible default though.

rickdott:
image
Results of the test, spawn indeed has a 'base' time added to it, but still grows nicely with dataset size.

rickdott:
Sorry to put another PCA commit here instead of in a new PR, just did it as the previous one was in here as well. The exact same output as before, just 4.5x faster (with 13k trials from 45s to 9s for just the PCA). Should be even more with bigger datasets.

GWeindel:
Sorry I did not finish my review, wanted to continue on the phone but I can't resume it. It all looks good from what I could see so far not sure the structure is estimator specific enough but if you're confident enough you can merge. We want that parallelization stable by 1.0.0 and we can rework on that before if needed

kiante-fernandez:
I found a problem with the chunked likelihood that I think needs addressing before this merges. The likelihood returned by a parallel fit depends on the number of cpus.

The cause is in how each chunk computes its pmf. `estim_probs` takes `max_duration` from the trials in its own subset, and `distribution_pdf` normalizes the pmf over that window with `p = p / np.sum(p)`. So a chunk that does not contain the longest trial normalizes over a shorter window than the serial computation, and its trial likelihoods come out different. With my test data below, about 22% of the pmf mass sits beyond the short chunk's window, so the effect is not just some float point noise.

Here is a minimal reproducer against the current branch head:

```python
import numpy as np
import xarray as xr

from hmp.models import EventModel
from hmp.patterndata import PatternData
from hmp.patterns import HalfSine

# 10 short trials (60 samples) then 10 long (200): np.array_split with 2
# chunks puts all short trials in chunk 0, whose local max_duration is 60.
durations_vals = np.array([60] * 10 + [200] * 10)
boundaries = durations_vals.cumsum()
starts = np.roll(boundaries, 1)
starts[0] = 0

durations = xr.DataArray(durations_vals, dims=("trial",),
                         coords={"trial": np.arange(len(durations_vals))})
pattern = HalfSine()
cross_corr = np.random.default_rng(0).normal(0, 0.3, size=(boundaries[-1], 3))

data = PatternData(durations=durations, starts=starts, ends=boundaries - 1,
                   sfreq=100.0, pattern=pattern, template=pattern.template,
                   cross_corr=cross_corr)

ref = None
for cpus in [1, 2, 3, 4, 5]:
    m = EventModel(n_events=2)
    m.fit(data, verbose=False, cpus=cpus)
    chunk_maxes = [int(durations_vals[c].max())
                   for c in np.array_split(np.arange(len(durations_vals)), cpus)]
    if ref is None:
        ref = m
    dtp = np.nanmax(np.abs(m.time_pars - ref.time_pars))
    print(f"cpus={cpus}  lkh={m.lkhs:.10f}  chunk max_durs={chunk_maxes}  "
          f"max|time_pars - cpus1|={dtp:.6f}")
```

Output on my machine:

```
cpus=1  lkh=-115.6418511154  chunk max_durs=[200]  max|time_pars - cpus1|=0.000000
cpus=2  lkh=-107.5727835367  chunk max_durs=[60, 200]  max|time_pars - cpus1|=0.203896
cpus=3  lkh=-109.9952627951  chunk max_durs=[60, 200, 200]  max|time_pars - cpus1|=0.203896
cpus=4  lkh=-107.5727835367  chunk max_durs=[60, 60, 200, 200]  max|time_pars - cpus1|=0.203896
cpus=5  lkh=-109.1877697089  chunk max_durs=[60, 60, 200, 200, 200]  max|time_pars - cpus1|=0.203896
```

Four different likelihoods from five cpu settings on identical data, and the fitted time_pars differ from the serial fit. You can see the likelihood track the chunk layout: it shifts by exactly how many short trials landed in a chunk whose local max is 60, which is why cpus=2 and cpus=4 agree with each other. The EM stopping rule takes in this likelihood, so tolerance checks, selection across starting points, and any model comparison downstream all become functions of cpu count.

`test_multiprocessing_matches_serial` does not catch this because its data is two trials of the same duration (71 samples), so every chunk's local max equals the global max and the comparison passes trivially. Making the durations in that test heterogeneous would turn it into a proper guard for this.

The good news is that the chunking itself seems fine. I think the fix is to compute max_duration once over the full subset in `chunked_estim_probs` and pass it into each per chunk `estim_probs` call. Test on your end and let me know what you think

rickdott:
Great catch! This fix outputs:
```
cpus=1  lkh=-115.6418511154  chunk max_durs=[200]  max|time_pars - cpus1|=0.000000
cpus=2  lkh=-115.6418511154  chunk max_durs=[60, 200]  max|time_pars - cpus1|=0.000000
cpus=3  lkh=-115.6418511154  chunk max_durs=[60, 200, 200]  max|time_pars - cpus1|=0.000000
cpus=4  lkh=-115.6418511154  chunk max_durs=[60, 60, 200, 200]  max|time_pars - cpus1|=0.000000
cpus=5  lkh=-115.6418511154  chunk max_durs=[60, 60, 200, 200, 200]  max|time_pars - cpus1|=0.000000
```
Which is looking better.

I'll give the estimator-specificity some more thinking before merging. I don't really like passing pool through so many functions, but having it as a class member makes pickling/multiprocessing more difficult.

PR #324, opened 2026-08-05
Sampling-based estimation with NUTS

kiante-fernandez:
Builds on #244, merge after that one.

Adds MCMCEstimator next to EMEstimator. Same likelihood, reimplemented in JAX so it can be differentiated, wrapped in a PyTensor Op with a hand-written VJP. It is registered as an observed distribution rather than added as a potential, which populates idata.log_likelihood, so the number of events can be picked by cross validation with a standard error instead of by walking up local maxima.

New modules: jax_likelihood.py (forward-backward), pytensor_op.py (PyMC wrapper, compiles to JAX under numpyro), mcmc.py (model and posterior summary). fit_likelihoods and max_explored_scale in base.py let the search strategies read likelihoods without touching EM internals; both take estimator= in their constructor. Sampling dependencies are an mcmc extra, and MCMCEstimator is imported lazily so hmp.estimators works without them.

Checks: likelihood matches numpy, per-trial Op matches the scalar one, gradient matches finite differences. Eight replicates on simulated data put 187/200 true values inside their 94% interval. LOO picks the true event count when used for model comparisons. On SAT speed it picks three events, as shown. All six tutorial operations run under both estimators. Suite at 132.

Not supported: any duration distribution but the gamma, and fixed_channel_pars/fixed_time_pars. Both raise rather than fit something else.

jitter defaults off. The perturbation is uniform on (-1, 1) unconstrained, about 2.7x on a log scale, and cold starts do not converge with it on. It also spreads the chains, which is what separated_modes reads, so turn it on when the question is multimodality. Need to think about this more.

The default backend runs one chain at a time: where multiprocessing forks its workers, JAX in the parent can deadlock them. numpyro runs chains in parallel. Another issue that will need more work.

An r hat near 1.8 here would come up sometimes; wrong event count, too few components, group differences that are not there. More draws do not help.

Priors are provisional. estim_probs gained an intermediates argument used only by the reference generator, worth a look in review.

### DIFF
--- a/hmp/basedata.py
+++ b/hmp/basedata.py
@@ -360,7 +360,7 @@ def from_io(epoch_data: xr.Dataset) -> BaseData:
     BaseData
         An instance of BaseData
     """
-    base_data = BaseData(data=epoch_data.copy())
+    base_data = BaseData(data=epoch_data)
     #process data from io
     base_data.data = base_data.data.data.stack(trial=["recording", "epoch"])\
         .dropna("trial", how="all")
--- a/hmp/estimators/em.py
+++ b/hmp/estimators/em.py
@@ -1,16 +1,31 @@
 """Expectation-Maximization estimator for HMP models."""
 
-import itertools
 import multiprocessing as mp
 from warnings import resetwarnings, warn
 
 import numpy as np
-from tqdm.auto import tqdm
 
 from hmp.patterndata import PatternData
+from hmp.utils import _get_mp_context
 
 from .base import BaseEstimator, EstimationResult
 
+_WORKER_DATA = {}
+
+def _init_worker(pattern_data: PatternData, model):
+    _WORKER_DATA["pattern_data"] = pattern_data
+    _WORKER_DATA["model"] = model
+
+
+def worker_estim_probs(channel_pars, time_pars, chunk, max_duration):
+    """Worker function to estimate probabilities of a chunk of trials."""
+    return _WORKER_DATA["model"].estim_probs(
+        _WORKER_DATA["pattern_data"],
+        channel_pars,
+        time_pars,
+        subset_epochs=chunk,
+        max_duration=max_duration,
+    )
 
 class EMEstimator(BaseEstimator):
     """Expectation-Maximization parameter estimator.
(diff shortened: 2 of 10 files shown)


kiante-fernandez:
To be clear about scope: this is a provisional first pass at getting an MCMC estimator working, not a finished feature. The diff looks large because it sits on top of #244 and GitHub is showing both until that one merges; the part that is actually new here is thirteen files.

Everything in it is what was needed to get the prototype running end to end, so some of it is groundwork rather than the sampler itself. Happy to split anything out if it is easier to review that way.

GWeindel:
I'll need quite some time to go through this but this seems like a perfect first take on estimating HMP using MCMC. I'm not so surprised on the jitter effect. I do think that for a first time the best option is to have a mixture between EM and MCMC at least to ensure that the chains do not get stuck in weird parameter space while still ensuring that it is explored appropriately. So yes definitely some thinking needed here.

I suggest we merge #244 once you accept https://github.com/kiante-fernandez/hmp/pull/1. Then I think this can stay a single PR. 

I'll be away from GitHub from Friday up to the 23rd of August but then I look forward to dig into this more deeply.

JoKra1:
This looks really cool, and I would be very happy with JAX as an extra dependency - this will make some of my work easier as well! 👍🏻 

One suggestion regarding efficiency @kiante-fernandez: when you only need the log-likelihood but not the event probs (as for NUTS) you do not need to evaluate the backward recursion for a HMP-like model. Rather than evaluating two convolutions in the loop, computing the log-likelihood only requires multiplying the last event forward variables by the flipped pmf of the last interval (see ``trial_likelihood2 `` below - this is for the basic event model which was easier for me to modify now since I have not checked your code in detail yet). The trial-likelihood is then the sum over the last event **not the first**!

```
for event in np.arange(
            1, n_events
        ):  # Following stage transitions integrate previous transitions
            add_b = backward[:, :, event - 1] * probs_b[:, :, event - 1]  # Next stage in back
            for trial in np.arange(n_trials):
                # convolution between gamma * gains at previous event and event
                forward[:, trial, event] = np.convolve(forward[:, trial, event - 1], pmf[:, event])[
                    :max_duration
                ]
                # same but backwards
                backward[:, trial, event] = np.convolve(add_b[:, trial], pmf_b[:, event])[
                    :max_duration
                ]
            forward[:, :, event] = forward[:, :, event] * probs[:, :, event]

        trial_likelihood2 = forward.copy()
        for trial in np.arange(n_trials):
            trial_likelihood2[: durations[trial], trial, -1] *= np.flip(pmf[: durations[trial], -1]) # Must flip only the part up to dur
        
        # re-arranging backward to the expected variable
        backward = backward[:, :, ::-1]  # undoes stage inversion
        for trial in np.arange(n_trials):  # Undoes sample inversion
            backward[: durations[trial], trial, :] = backward[: durations[trial], trial, :][::-1]
        eventprobs_raw = forward * backward
        trial_likelihood2 = np.clip(trial_likelihood2, 0, None) 
        eventprobs = np.clip(eventprobs_raw, 0, None)  # floating point precision error
        trial_likelihood = np.log(
            eventprobs[:, :, 0].sum(axis=0)
        )  # sum over max_samples to avoid 0s in log
        likelihood = np.sum(trial_likelihood)
        trial_likelihood2 = np.log(
            trial_likelihood2[:, :, -1].sum(axis=0) # MUST SUM OVER LAST EVENT!
        )
        print(likelihood, np.sum(trial_likelihood2))
```

``likelihood`` and ``np.sum(trial_likelihood2)`` are equivalent up to rounding error occurring during convolution of backward (so ``np.sum(trial_likelihood2)``) should even be more stable.

PR #327, opened 2026-08-07
Decreased RAM usage for read_mne_epochs

rickdott:
Encountered an issue that the 128GB RAM I have access to filled up when reading a 35GB dataset using `read_mne_epochs()`

Currently, the function keeps both the `epochs` and the `hmp_data_format()` in memory until the list comprehension finishes, requiring at least 2x the dataset size in RAM.

This fix iteratively removes the already converted epochs from memory while building the list of xr.Datasets.

### DIFF
--- a/hmp/io/mne_epochs.py
+++ b/hmp/io/mne_epochs.py
@@ -109,28 +109,32 @@ def read_mne_epochs(
                 ],
             )
 
-    epoch_data = [
-        utils.hmp_data_format(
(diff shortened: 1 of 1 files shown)


rickdott:
I saw RAM usage going up to 78GB for this dataset while populating during the first step, at which it would need twice that (156GB) for the list comprehension.

I still think 78GB is a bit much for a 35GB dataset, but this is workable at least.

